PR #7, opened 2015-05-10
Added premake script for project file generation

CaptainZippy:
This allows generating project files for msvc (tested) and other perhaps other platforms (not tested).

I'll open a separate ticket for msvc issues.
See http://premake.github.io/ for more info about premake

### DIFF
--- a/premake5.lua
+++ b/premake5.lua
@@ -16,29 +16,36 @@ solution "PEGTL"
       optimize "On"
 
 project "pegtl"
+    kind "None"
     files { "pegtl/**.hh" }
 
-group "Test"
-project "UnitTest"
-    files { "unit_tests/**.hh", "unit_tests/**.cc" }
-    pchheader "test.hh"
-    pchsource "unit_tests/trace.cc"
+group "UnitTest"    
+    for _,proj in next,os.matchfiles("unit_tests/*.cc"),nil do
+        for name in string.gmatch(proj, "/([%a%d_]+)\.") do
+            project(name)
+            files { proj }
+        end
+    end
 
-examples = {
-    "calculator",
-    "hello_world",
-    "json_build_one",
-    "json_build_two",
-    "json_parse",
-    "lua53_parse",
-    "modulus_match",
-    "s_expression",
-    "sum",
-    "unescape",
-    "uri_trace"
-    }
+    project "include"
+        kind "None"
+        files { "unit_tests/**.hh" }
+
 group ("Examples")
-for _,proj in next,examples,nil do
-    project (proj)
-       files { "examples/" .. proj .. ".cc" }
-end
+    examples = {
+        "calculator",
+        "hello_world",
+        "json_build_one",
+        "json_build_two",
+        "json_parse",
+        "lua53_parse",
+        "modulus_match",
+        "s_expression",
+        "sum",
+        "unescape",
+        "uri_trace"
+    }
+    for _,proj in next,examples,nil do
+        project (proj)
+           files { "examples/" .. proj .. ".cc" }
+    end
--- a/premake5.lua
+++ b/premake5.lua
@@ -1,0 +1,44 @@
+-- premake5.lua
+
+solution "PEGTL"
+    location "build"
+    includedirs { "." }
+    kind "ConsoleApp"
+    language "C++"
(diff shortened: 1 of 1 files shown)


ColinH:
Thank you for this contribution, we are aware that the PEGTL is still missing something to build out-of-the-box on Windows; not using Windows ourselves we are a bit unsure regarding what the best practice would be, and of course we can't test anything.

Sam, who helped us with VS2015 compatibility, has a forked repository where he checked in some VS project files, which, compared to the premake approach, seems to have both advantages and disadvantages:

The finished project files seem to depend on the exact set of files in the release archive and would therefore have to track the project development, but:

With premake people need to install another tool in order to get things running, which somehow defeats the idea of out-of-the-box support -- and then the question is whether to use premake, CMake, ...

Also, should we use a platform-independent build system it would be nice to use it for all platforms, including Unix and Xcode, but without making things more complicated.

What do you think from your point-of-view as VS user, regarding future-proofing, simplicity and best practice?

kirbyfan64:
@ColinH Having used VS before...why not use something like [Bakefile](http://bakefile.org/)? It generates portable VS project files that you can check in to source control, along with makefiles.

As a result, you could maybe use Bakefile for both, which would avoid having the VS project files go out of sync with the makefiles.

CaptainZippy:
I really like /some/ kind of autogeneration of project files, especially since, as you say checked-in project files can go out of date quite easily. Annoyingly, checked-in project files often need the IDE installed to update them.

I've been very impressed with what I've seen of premake. It feels very lightweight (just one executable) in comparison to other systems such as cmake. I haven't tried bakefile, but it seems to support less platforms. I'll note that premake would cover windows, mac and could even replace handwritten makefiles if needed. In fact, I'd recommend dropping the hand-written makefiles in favour of generating them also.

One approach I'd suggest is that the platform specific project files can be created on-the-fly when packaging for release. They'd get zipped up into an extra build/ folder. This is what premake itself does. Or they could be checked in by a bot, though I wouldn't like the extra checkins.

d-frey:
I was wondering about that last paragraph for quite some time: How do I add files on-the-fly when "packaging" for release? It seems to me that GitHub only tags the repository, there is no real script to use some checked-in file (for premake, CMake, etc.) and *replace* it with the generated files for VS, XCode, GNU Make. I agree that this would be the right approach. You seem to suggest it is possible, any link would be appreciated.

And while we are at it: How do I make GitHub add the (pre-rendered) Wiki pages to the release archive? Again, any help appreciated!

kirbyfan64:
Wow, did *no* one read my comment? ;)

d-frey:
@kirbyfan64 I did, but I haven't found the time to try it out. Also, if you already know the tool, it would be very helpful if you could put together a bakefile for the PEGTL. Even if it is not perfect, it would be much easier to get started and evaluate it from my end. Our current Makefile is not too complicated, feature parity should be easy, shouldn't it? ;)

kirbyfan64:
@d-frey I'll try!

samhocevar:
Well, since the PEGTL is a header-only library, users will use whatever build system their project has and never actually use yours, except maybe to try the samples, so be careful about not driving users away if they have to install additional software just to try things out.

@kirbyfan64 note that Bakefile doesn’t support Visual Studio 2015 yet (but it shouldn’t require much work).

kirbyfan64:
@samhocevar vslavik/bakefile#55 I know. :)

CaptainZippy:
@d-frey - I'm not familiar with github-specific release issues. When releasing my project (way back on sourceforge!), I used to tag, then run a script locally to build packages and upload them to various sites.

Github seems to allow attaching extra things to a release: https://github.com/blog/1547-release-your-software No doubt there's some api so one could automate this part.

d-frey:
We decided to close this pull request and the whole topic of providing VS project files for now. The different build tools (CMake, premake, Bakefiles, ...) are all (AFAIK) not pre-installed, if we generate the files needed for VS, we would need to test/update them for all versions of VS and they are still too complicated for what we actually need.

After all, the PEGTL is a small library, the Makefile for Unix/MacOS is just 66 lines. VS needs something similar, else the overhead and complexity is in our opinion too high. As the PEGTL is a header-only library, it should also be easy enough for interested users to integrate it into their projects manually.

PR #11, opened 2015-12-10
Support for Android

kneth:
Android's C++ library lacks `std::to_string()` and `errno` must be included using the correct header file.


d-frey:
The patch does not compile (but that is easy to fix). Also, I'd like to avoid conditional compilation is possible and both you changes can be used in general, so there is no need to have Android-specific #ifdefs. Can you confirm that with those two things fixed the rest of the PEGTL is supported on Android? (All tests pass, ...)

kneth:
In order to run the unit tests on an Android device, we will have to write a native activity (http://developer.android.com/ndk/samples/sample_na.html). I don't mind doing that, and I will have some time in the weekend to do it.

d-frey:
Thanks, let me know when the tests on Android pass.

d-frey:
@kneth Have you found the time to run the tests on Android?

kneth:
@d-frey I spent a few hours on it last weekend, and did some progress. The unit tests are compiled and executed one-by-one which is really not a good idea when it comes to Android: generating a native app per unit tests and executing it on a device/emulator will be cumbersome. I hope to have a little time to spare during the xmas break.

kneth:
@d-frey I have added some C++ code as a driver for running a single unit test under Android. Furthermore, I have added a couple of shell scripts to help modifying the unit test and running it on an Android device. I will add a proper README tomorrow so it will become clear how to use it.

kneth:
Current status: `contrib_json.cc` crashes in `pegtl::mmap_parser::mmap_parser`.

d-frey:
Kenneth, I committed two changes without the conditional compilation guards. Please update your repo/patch.

Since I'm not an Android developer, I'd like to see some automated tests to make sure we won't break Android with future changes. Could you maybe take the time and modify you repository's .travis.yml to include automated tests? If possible for all applicable Android versions? Here's what I found from TravisCI: https://docs.travis-ci.com/user/languages/android

Once those automated tests pass (and the remaining mmap-issue is solved or mmap is disabled for Android if it's not possible/allowed), I'd like to add the rest to the main repo.

Thanks for your work so far!

kneth:
@d-frey Thanks for the update.

I'll take a look at Travis to see if I can get it to work.

d-frey:
@kneth Hi Kenneth, any update regarding Travis?

ColinH:
Most of this pull request should now be integrated into the main branch, if in slightly different form. For now I will close this request, further work on Android should be done separately. It would be great if you - or somebody else more familiar with Android development than ourselves - could confirm that everything works, or help fix any remaining issues...

seniorivn:
So it's not working on Android? Am i right?
Does anyone know library for peg that can work on android and ios?

d-frey:
@seniorivn I hope I fixed all the issues Kenneth reported, so I it should work - unless there are more issues no one reported yet.

What we don't have right now is automated testing. So, first of all you could simply test and see if it works for you on Android. If it does, please let us know. If it does not, please let us know *why* it does not work. And if you'd really like to help, help us to set up automated testing for Android :-)

seniorivn:
@d-frey thanks, good news.
If i'll start using it, i'll help, at least by telling you how it works for me.

PR #18, opened 2016-04-04
Fix to compile and pass all tests in MinGW on Windows

pauloscustodio:
The gcc 4.9.3 include files on MinGW guard all non-Ansi functions with '#ifndef __STRICT_ANSI__', causing compilation errors in functions like strcasecmp(); fix by adding '-U__STRICT_ANSI__' option if
'gcc -dumpmachine' returns 'mingw32'.

internal::file_mapper and internal::file_opener are only defined if mmap() exists; guard the corresponding tests with '#if defined(_POSIX_MAPPED_FILES)'.

verify_file assumes that the data input file has Unix line endings, which is not true on Win32 and causes test failures; fix by modifying the grammar to include:
  struct any_eol : sor< one< '\n' >, pegtl::string< '\r', '\n' >,	one< '\r' > > {};

Maybe any_eol could be included in ascci.hh?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,9 @@
 
 ifeq ($(OS),Windows_NT)
 UNAME_S := $(OS)
+ifeq ($(shell gcc -dumpmachine),mingw32)
+MINGW_CXXFLAGS = -U__STRICT_ANSI__
+endif
 else
 UNAME_S := $(shell uname -s)
 endif
@@ -23,7 +26,7 @@ endif
 # changed if desired.
 
 PEGTL_CPPFLAGS ?= -pedantic
-PEGTL_CXXFLAGS ?= -Wall -Wextra -Werror -O3
+PEGTL_CXXFLAGS ?= -Wall -Wextra -Werror -O3 $(MINGW_CXXFLAGS)
 
 .PHONY: all clean
 
--- a/unit_tests/internal_file_mapper.cc
+++ b/unit_tests/internal_file_mapper.cc
@@ -7,6 +7,7 @@ namespace pegtl
 {
    void unit_test()
    {
+#if defined(_POSIX_MAPPED_FILES)
       try {
          internal::file_mapper( "pegtl" );
          std::cerr << "pegtl: unit test failed for [ internal::file_mapper ] " << std::endl;
@@ -14,6 +15,7 @@ namespace pegtl
       }
       catch ( const std::exception & ) {
       }
+#endif
    }
 
 } // pegtl
--- a/unit_tests/internal_file_opener.cc
+++ b/unit_tests/internal_file_opener.cc
@@ -7,6 +7,7 @@ namespace pegtl
 {
    void unit_test()
    {
+#if defined(_POSIX_MAPPED_FILES)
       const internal::file_opener fo( "Makefile" );
       ::close( fo.m_fd );  // Provoke exception, nobody would normally do this.
       try {
@@ -16,6 +17,7 @@ namespace pegtl
       }
       catch ( const std::exception & ) {
       }
+#endif
    }
 
 } // pegtl
--- a/unit_tests/verify_file.hh
+++ b/unit_tests/verify_file.hh
@@ -8,7 +8,8 @@
 
 namespace pegtl
 {
-   struct file_content : pegtl_string_t( "dummy content\n" ) {};
+   struct any_eol : sor< one< '\n' >, pegtl::string< '\r', '\n' >,	one< '\r' > > {};
+   struct file_content : seq< pegtl_string_t( "dummy content" ), any_eol > {};
    struct file_grammar : seq< rep_min_max< 11, 11, file_content >, eof > {};
 
    template< typename Rule > struct file_action : nothing< Rule > {};


ColinH:
Let's see, those are three separate issues:

1. The strict ansi business seems to be a bug in MinGW that might however require a workaround until they get it fixed.
2. The headers for `internal::file_mapper` and `internal::file_opener` should only be included by `file_parser.hh` when the platform supports `mmap()`, they don't guard themselves. Did we miss something?
3. We recently changed the `read_parser` to open the files with mode "rb" instead of "r", could you please check whether that fixes the issue?

(Independently we might add some further convenience rules for different line-ending conventions, however our unit test uses our Makefile, and that has Unix line endings.)

pauloscustodio:
Answer for each of the questions:

1. Correct. The -U__STRICT_ANSI__ is a workaround required to be able to compile PEGTL in mingw32. For this reason it is only included in the compile options if 'gcc -dumpmachine' returns 'mingw32'.

2. Without the guards in the pull request, the compilation of unit_tests/internal_file_mapper.cc and unit_tests/internal_file_opener.cc fail:  
   ```
g++ -std=c++11 -I. -pedantic -Wall -Wextra -Werror -O3 -U__STRICT_ANSI__ unit_tests/internal_file_mapper.cc -o build/unit_tests/internal_file_mapper
unit_tests/internal_file_mapper.cc: In function 'void pegtl::unit_test()':
unit_tests/internal_file_mapper.cc:11:10: error: 'file_mapper' is not a member of 'pegtl::internal'
          internal::file_mapper( "pegtl" );
g++ -std=c++11 -I. -pedantic -Wall -Wextra -Werror -O3 -U__STRICT_ANSI__ unit_tests/internal_file_opener.cc -o build/unit_tests/internal_file_opener
unit_tests/internal_file_opener.cc: In function 'void pegtl::unit_test()':
unit_tests/internal_file_opener.cc:10:23: error: 'file_opener' in namespace 'pegtl::internal' does not name a type
       const internal::file_opener fo( "Makefile" );
unit_tests/internal_file_opener.cc:11:7: error: '::close' has not been declared
       ::close( fo.m_fd );  // Provoke exception, nobody would normally do this.
unit_tests/internal_file_opener.cc:11:16: error: 'fo' was not declared in this scope
       ::close( fo.m_fd );  // Provoke exception, nobody would normally do this.
   ```
3. Without the patch in the pull request, test build/unit_tests/file_file fails:  
   
   ```
build/unit_tests/file_file
pegtl: unit test assert [ p.template parse< file_grammar >() ] failed in line [ 49 ] file [ unit_tests/verify_file.hh ]
pegtl: unit test assert [ p.template parse< file_grammar >( flag ) ] failed in line [ 56 ] file [ unit_tests/verify_file.hh ]
pegtl: unit test assert [ p.template parse< file_grammar >( flag ) ] failed in line [ 63 ] file [ unit_tests/verify_file.hh ]
pegtl: unit test assert [ result ] failed in line [ 71 ] file [ unit_tests/verify_file.hh ]
pegtl: unit test assert [ flag == true ] failed in line [ 72 ] file [ unit_tests/verify_file.hh ]
pegtl: unit test assert [ result ] failed in line [ 79 ] file [ unit_tests/verify_file.hh ]
pegtl: unit test assert [ flag == true ] failed in line [ 80 ] file [ unit_tests/verify_file.hh ]
   ```
   
   Changing the open option to "r" in pegtl/internal/file_reader.hh causes an input_error exception, as pegtl::internal::file_reader() expects to read 165 bytes from the input, the file size with Win32 line endings, and only reads 154 bytes, as each "\r\n" is replaced by "\n" by the text read layer in the OS.   

   Other tests fail for the same reason: the number of read bytes is not the same as the file size. 

   Changing the open option to "r" also prevents PEGTL to be used to parse binary files on Windows.

   In my opinion, the option of opening the file in binary mode, i.e. without any OS translation, and generalize the eol and eolf parsers to accept the common line endings ("\n", "\r\n", "\r") is more general. 

4. Although the unit_tests/file_data.txt is stored in the repository with Unix line endings, these are changed by git to the host line endings when pulling, and back to Unix when pushing.

ColinH:
4. Never having used git on Windows I wasn't aware of this.

3. I have changed the `eol` and `eolf` rules to accept both Unix (`\n`) and Dos (`\r\n`) line endings.

2. Right, this is now fixed.

1. This change is also included.

I didn't merge your pull request because I wanted to handle 3. differently.

If you could verify that I got everything right...

Thanks!

pauloscustodio:
Almost right! The test unit_tests/file_file.cc was still failing because of the hard-coded literal "\n" instead of eol in the rule.

Fixed and issued a new pull-request for it.

A single "\r" is also used in some (older) architectures as line end character, as the  Mac OS up to version 9, and OS-9, and should also be accepted as eol character, if the other two rules "\n" and "\r\n" fail. 

Thanks!

ColinH:
Yes, thanks, guess it shows that we can't currently test on Windows ourselves.

We thought about the "\r", but for the moment we have decided to bet on it being obsolete.

Of course if somebody starts parsing a bunch of old files with "\r" we will reconsider...

ColinH:
Btw, which name can I use to include you in the "Thank You" section? Is it "Paulo S. Custodio", or "Paulo Scustodio", or am I guessing completely wrong?

pauloscustodio:
Paulo Custodio is fine.

I do parse old files with "\r" line endings and would like to have eol
handling it instead of creating my own specialization. But of course it is
your call.

It is a pity that Travis CI does not support Win32.

Thanks!

Regards,
Paulo Custodio

ColinH:
Thank you!

I'll keep the "\r" in mind. To support it properly the input class needs changing, too. It keeps track of the current line number by counting "\n"s...

PR #21, opened 2016-04-14
Support Mac OS style line ending

kuzmas:
OS X style line ending is just '\r'.
This patch adds support for this line ending style.

### DIFF
--- a/pegtl/internal/eol.hh
+++ b/pegtl/internal/eol.hh
@@ -33,8 +33,13 @@ namespace pegtl
                in.bump_next_line();
                return true;
             }
-            if ( ( a == '\r' ) && ( s > 1 ) && ( in.peek_char( 1 ) == '\n' ) ) {
-               in.bump_next_line( 2 );
+            if ( a == '\r' ) {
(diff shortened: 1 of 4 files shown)


ColinH:
Actually "\r" is Mac OS up to OS 9. Mac OS X is Unix-based it uses Unix line ending, "\n". That said, you are not the first person to ask for "\r"-support. We will check your patch, thank you!

ColinH:
Changing the `eol`-rule is not enough, the `input`-classes count `\n`-characters to determine the current line number, which is used for `parse_error` exceptions. I wanted to check out another, unrelated change regarding the line number counting and will try to add support for `\r`, too.

kuzmas:
Oh, yeah, you are absolutely right - it's OS 9, not OS X. It just happened that I use OS X at work and we still have a lot of files with "\r" in our projects.

ColinH:
This will take some more work to do in a way that reported line numbers are always correct. I opened #22 as reminder and will close this pull-request.

kuzmas:
Ok

PR #28, opened 2016-07-14
Fix #27.

### LINKED ISSUE #27: "Hello, World!" example does not seem to work
`pegtl::input` does not seem to exist.

Axure:


ColinH:
The renaming of the input classes was done after 1.3.x so the documentation is correct in that branch.

PR #31, opened 2016-09-24
Expose raw byte position in position information.

samhocevar:
In some cases it may be desirable to know the byte offset in addition
to the line / column information, for instance when parsing binary
files, or when feeding the parser with partial data.

### DIFF
--- a/pegtl/action_input.hh
+++ b/pegtl/action_input.hh
@@ -15,8 +15,8 @@ namespace pegtl
    class action_input
    {
    public:
-      action_input( const std::size_t in_line, const std::size_t in_byte_in_line, const char * in_begin, const char * in_end, const char * in_source )
-            : m_data( in_line, in_byte_in_line, in_begin, in_end, in_source )
+      action_input( const std::size_t in_byte, const std::size_t in_line, const std::size_t in_byte_in_line, const char * in_begin, const char * in_end, const char * in_source )
+            : m_data( in_byte, in_line, in_byte_in_line, in_begin, in_end, in_source )
       { }
 
       bool empty() const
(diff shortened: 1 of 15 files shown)


samhocevar:
This commit changes the API so I am not sure it is acceptable
as is. However the feature is desirable for my purposes, and I
found out it was easier to modify the PEGTL rather than implement
a byte tracking mechanism using additional state variables, which
I would then have to add to every parser needing it.

I am also not sure of how to format the information (or whether I should
ignore it) in `position_info::operator<<`.

If this does not fit within your overall plans, maybe you can suggest a
more elegant solution?

ColinH:
Thanks for the pull request, I'll not merge it right now because it might interfere with some other things that we have planned, but keep it as reminder and for possible future inclusion if sufficiently independent of these other changes.

ColinH:
Do you later use the byte position programmatically, or do you only need the human-readable form in the exception message?

samhocevar:
I use the byte position programmatically, yes.

My current use case here is the creation of a transpiler that does not need an intermediate AST. The parser analyses the input and marks parts of the code using their byte offsets. A post-process then uses search/replace to perform the language transformation, and if the replaced chunk changes the size, offsets located after it get updated. Using line/column notation would make the search/replace work more complex.

ColinH:
Ok, thanks, another small question: Do you ever need the byte offset together with column and line, or do you not need the column and line in cases where you use the byte offset?

samhocevar:
In my case, unless I am parsing binary data, I think I always need column and line in order to provide meaningful error reporting, regardless of whether my action class uses the byte offset.

ColinH:
I've been trying to make this more flexible, to not always keep all three numbers, but it seems to be more work than anticipated, and only for a small optimisation. Your pull-request is now unfortunately out of date, but some of the changes will be simpler with the latest commits. I will now look into re-doing this.

PR #36, opened 2016-12-06
Fix EOL handling

kuzmas:
Fix for issue mentioned in https://github.com/ColinH/PEGTL/issues/22#issuecomment-265087905


d-frey:
Nice catch and fix, thanks!

PR #37, opened 2016-12-06
Generalized action input

kuzmas:
Some inputs are not char-based, therefore basic_action_input cannot be used.
This change provides a more general approach while keeping the default behaviour intact.

In our project for performance reasons we use multi-pass parsing. First pass is a usual char-based lexer pass which splits input file into tokens, then this token stream is fed to the second pass, to token parser.

This change is needed for this second pass parser.

### DIFF
--- a/pegtl/action_input.hh
+++ b/pegtl/action_input.hh
@@ -18,6 +18,11 @@ namespace pegtl
    class basic_action_input
    {
    public:
+
+      basic_action_input( const std::size_t in_line, const std::size_t in_byte_in_line, const char * in_begin, const char * in_end, const char * in_source )
+            : m_data( in_line, in_byte_in_line, in_begin, in_end, in_source )
+      { }
+
       basic_action_input( const internal::input_mark& from, const internal::input_mark& to )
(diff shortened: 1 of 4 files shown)


ColinH:
That's interesting, we had thought about something similar a while ago, is it possible to see more of the source and how everything fits together in your solution?

Is there a particular reason why the new constructor for class `action_input` takes two `input_mark`s

I'm asking because `rule_match_two` uses data from the `input_mark` and the `input` to create the `action_input`, and I was just working on something else that also required calling the constructor of `action_input` with the `input_mark` and the `input` instead of having `rule_match_two` extract the fields, and if that works for your case, too, then we could still make do with a single constructor...

kuzmas:
It works fine, thanks.

We actually do not use the second argument to action_t constructor anyway, so exact signature does not really matter.

As to your request to see more of our source code - it's actually a proprietary code of the company I work for, so I'll have to ask for authorisation.

kuzmas:
Sorry, no authorisation granted - our company does not allow to open the source code.

ColinH:
I understand, thanks for trying!

PR #43, opened 2017-01-17
Fix most of the warnings under MSVC.

lichray:


ColinH:
Thanks for this! Looking at the changes, some seem to be generic improvement, while others are very specific code uglifications to shut the compiler up, which is fine as long as small.
Two remarks, if `fopen_s()` is in the `std::` namespace, could you please write `std::fopen_s()` (or did MS not put it into that namespace)?
And in `file_reader`, if `fopen_s()` fails then `errno` is not set correctly, assuming that `errno` is thread-safe on Windows, can you assign to `errno` the return value of `fopen_s()` in or before the `if`-statement?

lichray:
`fopen_s` is C-only, at best I can qualify it with `::`, but I think this name is unique enough.  What do you think?
`fopen_s` sets `errno` *and* returns it (https://msdn.microsoft.com/en-us/library/z5hh6ee9.aspx), so there is no need to do it again.
One more thing, actually I can silence all warning by using VS 2015 update 2's new compiler option `/utf-8`, should I do so?  (Appveyor uses update 3 so that's fine, but users who do not update their compilers may get confused.)

ColinH:
Ok, since C++ puts other C-standard-library-functions into `std::` I was assuming that `fopen_s()` would be there, too, but then again it's not a **standard** library function.
Sorry, my bad, if `errno` is also set then all is well.
You can silence all of the warnings that you fix here with a `/utf-8` compiler option? That sounds strange, can you link to a documentation of that switch?

lichray:
`fopen_s` is a part of C11 standard (Annex K, optional feature), but C++ intentionally excludes it when updating the C reference to C11.

On `/utf-8`: https://blogs.msdn.microsoft.com/vcblog/2016/02/22/new-options-for-managing-character-sets-in-the-microsoft-cc-compiler/ because the last one or two warnings are caused by the environment not the code.  By default MSVC assume the source files uses your locale's code page, and Appveyor's locale has a narrow code page windows 1252.  The `/utf-8` option will force the compiler to use UTF-8 as source charset and execution charset, so that compiler will stop warn about having a Unicode character (even escaped) in literals that is not representable in the execution charset.

d-frey:
Also from my side: Thank you. I will clean up some parts of your patch later but in general, removing warnings for VS is great, we might even be able to add the `-Werror` equivalent to the build soon :)

d-frey:
Good news: After the cleanup and adding `/utf-8`, I was also able to add `/WX` so we compile on Visual Studio 2015 without a single warning now.

PR #49, opened 2017-02-01
Support MSVC with partial C++11 support

kuzmas:
MSVC before VS2015 do not support some of C++11 features, namely:
- noexcept
- constexpr
- inline namespace

This patch applies a workaround to this issue.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,6 +1,8 @@
 version: '{branch}-{build}'
 
-os: Visual Studio 2015
+os:
+  - Visual Studio 2013
+  - Visual Studio 2015
 
 configuration: Release
 
--- a/pegtl/analysis/insert_guard.hh
+++ b/pegtl/analysis/insert_guard.hh
@@ -16,7 +16,7 @@ namespace PEGTL_NAMESPACE
       class insert_guard
       {
       public:
-         insert_guard( insert_guard && other ) noexcept
+         insert_guard( insert_guard && other ) PEGTL_NOEXCEPT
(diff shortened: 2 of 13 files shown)


d-frey:
We are certainly interested in this, but I'd like to clarify: Is this *all* that is needed? Which versions of VS will be supported with this?

And finally: Could you please extend the appveyor.yml to addd automated tests for earlier versions of VS? Without the latter, it is too easy for us to break it again in the future and I'd like to avoid that. In fact, this would be my top priority before I'd consider merging it.

One last comment: I'd drop the macros for the inline namespace and replace it directly for all versions, as it seems functionally equivalent.

kuzmas:
We have to use MSVC from VS2013 and this is all I had to change to make it compile.

I have found another usage of `constexpr` in examples/csv2.cc - but I guess it's fine as it is.

kuzmas:
Well, it looks like we here were lucky enough to use only a small subset of PEGTL and managed to compile with VS2013. 
But when I tried to tackle the unit tests...  the incredible amount of them failed.

So my conclusion is: it's pretty safe to say that the VS2013 is not supported.

End of story ;-(

d-frey:
Yeah, I was afraid that this is the case. We are currently using high warning levels on appveyor which you could try to lower in the beginning, but it seems there are also a lot of really hard errors which can not be solved easily. Mostly with detecting overload by SFINAE, but also others that leave me clueless. Sorry it didn't work out, but thanks for trying!

PR #53, opened 2017-03-27
doozer: Use Doozer's built-in jobserver for make.

andoma:
Just call make like 'make -j' for this to happen. When scheduling multiple
jobs on the same agent Doozer can act as jobserver for all 'make' jobs
and distribute load evenly between multiple parallel jobs.

### DIFF
--- a/.doozer.json
+++ b/.doozer.json
@@ -3,72 +3,72 @@
     "jessie-i386": {
       "buildenv": "jessie-i386",
       "builddeps": ["build-essential"],
-      "buildcmd": "make -j ${PARALLEL}"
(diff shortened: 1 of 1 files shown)


d-frey:
OK, let's give it a try, but I have my doubts as especially for template-heavy code like the PEGTL the problem isn't CPU usage - it's memory usage. Once you start too many compile-processes in parallel you'll quickly run into swap-space and into a stalled system.

d-frey:
Looks like Doozer.io provides enough memory, and the build seems to work quite well with "make -j". Thank you, @andoma !

PR #59, opened 2017-05-29
Add Conan support, because Conan is super awesome

DEGoodmanWilson:
If you haven't seen (Conan)[https://www.conan.io], I recommend having a look. It's a package manager for C++. This PR adds a `conanfile..py`, which is necessary for pushing PEGTL to Conan—you would also need to install the Conan tool to release on Conan, of course.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -1,0 +1,19 @@
+from conans import ConanFile, CMake, tools
+import os
+
+
+class PEGTLConan(ConanFile):
+    name = "PEGTL"
+    version = "2.0.0"
+    license = "MIT"
+    username = "taocpp"
+    url = "https://github.com/%s/PEGTL.git" % username
+
+    def source(self):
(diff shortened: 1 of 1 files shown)


DEGoodmanWilson:
(You can check that this works for yourself by having a look at the testing release I've made `PEGTL/2.0.0@DEGoodmanWilson/testing` )

ColinH:
Thanks for the suggestion. While we are open to having the PEGTL included in as many package managers as reasonable, we cannot help but notice that none of the other packages made from the PEGTL require us to include configuration files, or even tools, to support the package system in question. Is there a non-invasive way of creating a PEGTL package for conan? We support package maintainers, but are not package maintainers ourselves - there are simply too many.

DEGoodmanWilson:
Thanks for the kind response. In fact, managing the package externally and in a non-invasive way is something that can be done easily: I just wanted to offer this first. No problem!

PR #65, opened 2017-07-19
Add utf8::bom and utf16::bom helper rules.

samhocevar:
I find myself reimplementing a rule for the optional [byte-order mark](https://en.wikipedia.org/wiki/Byte_order_mark) in every UTF-8 parser I write. The character is automatically added by many Windows-based text editors, so it makes sense to have it in `pegtl::utf8` and `pegtl::utf16`.


d-frey:
Thanks, `bom` seems to be sufficiently general and is a welcomed addition. I also added a `bom` rule for UTF-32 to detect them, plus some cleanups.

PR #67, opened 2017-08-04
Fix bug in Lua table parsing.

samhocevar:
The parser was unable to parse “{a==0}” because it believed “a=” was the
beginning of a table constructor field initializer. This is fixed by
disallowing “==” when matching “=”.


ColinH:
Thanks!

PR #70, opened 2017-09-05
Make sure we are not accessing a non-existent child.

joto:
This checks that `children` is not empty so calling `c.back()` later can not fail. Otherwise we crash for something like `1+2+`.


d-frey:
Thanks for spotting this problem. I'll commit a different fix, as you are just trying to handle a situation that should not occur in the first place. The root of the problem is that I do not clean up the stack properly in case of local failure. Doing so would probably be advisable in a real-world scenario, for this example I think it should be enough to convert some local failures into global failures. That way the child nodes will never be empty and errors are reported properly via an exception.

PR #74, opened 2017-11-03
Improve android build in cmake

Bjoe:
As I told on the C++ user group NRW meeting, I have some idea to improve the android build in travis and cleanup the cmake build system.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -163,38 +163,41 @@ matrix:
         - CXX=clang++
 
     - language: android
-      jdk: openjdk7
+      jdk: openjdk8
       android:
         components:
+          - tools
+          - platform-tools
           - tools
           - build-tools-23.0.3
           - extra-android-support
           - extra-android-m2repository
           - android-22
           - sys-img-armeabi-v7a-android-22
-          - cmake
+          # Unfortunaly travis use the old android sdk tool instead of using the new sdkmanager.
+          # In this deprecated android sdk tool there is no ndk-bundle or cmake package!
         licenses:
           - '.+'
+      before_install:
+        # install NDK and cmake via android sdkmanager. Unfortunaly to acknowledge the licenses 
+        # takes a while. Looks like the "echo y" solution is not working. Needs improvement !
+        - export TERM=dumb
+        - echo "y" | /usr/local/android-sdk/tools/bin/sdkmanager "ndk-bundle" "cmake;3.6.4111459"      
       before_script:
         - export TERM=dumb
-        - curl -L http://dl.google.com/android/repository/android-ndk-r10e-linux-x86_64.zip -O
-        - unzip -q android-ndk-r10e-linux-x86_64.zip
-        - export ANDROID_NDK_HOME=`pwd`/android-ndk-r10e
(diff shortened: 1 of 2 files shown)


d-frey:
Highly appreciated, thanks!

As it currently fails, I suggest you temporarily comment out the other builds in the `.travis.yml` to get faster feedback and safe some resources. Just re-enable them once your patch is working well.

I also have a question about the version used: Am I correct to read this as an upgrade in the tested Android version? Could your way be used to test different versions in a reasonable way? Our goal is to ensure that "reasonably old" versions of Android are supported as well as newer ones. (Just like with GCC/Clang/macOS/...) If that could be taken into consideration, that would be great :)

Bjoe:
@d-frey I'm done, it works now  :-)

Unfortunately the android built it takes ~10min :-( It takes to wait if the emulator is ready and to acknowledge the license ... For the license it looks like "echo y" is not working. I have no clue yet to fix this.

> I also have a question about the version used: Am I correct to read this as an upgrade in the tested Android version?

What you mean with the Android version? API version? Or SDK ? Compiler?...

d-frey:
Thank you very much!

I'll clean up some small stuff, I don't think it will make a real difference. Except maybe one thing: As the TravisCI build slaves are sometimes really slow, the `android-wait-for-emulator || android-wait-for-emulator` is actually needed or else you'd get some broken builds from time to time.

For the question about the versions: I don't know which versions of Android are still relevant / in use. Would it be possible to have multiple build jobs, one with, say, Android 22 (API level 22??), another one with Android 23, etc.? Does this make any sense? (I'm totally lost with all those API-level, NDK, etc. version numbers in Android)

Bjoe:
I'm also not an android expert and I do not even have a android phone (only BB10) but in my company we worked with android (also with iPhone and BB10) so I know a little bit and Yes: This Android 22 is API level 22 ... Android 19 is API level 19 ... 
You can read more about on [What is API Level](https://developer.android.com/guide/topics/manifest/uses-sdk-element.html#ApiLevels) and [Android NDK Native APIs](https://developer.android.com/ndk/guides/stable_apis.html)

At the moment your project is only building for Android API level 22. If you plan to have another API level, for example API level 19, you must change in travis.yml following lines:
```
components:
  ...
  - android-19
  - sys-img-armeabi-v7a-android-19
  ...
 script:
    ...n/cmake .... -DANDROID_PLATFORM=android-19 ....
```
That's it.

d-frey:
I tried that and it worked for Android 19.

It didn't work for Android 23 and 26, it complains about the `--abi`. I tried `arm64-v8a` but that didn't work either. Do you know how to fix this? Thanks.

Bjoe:
@d-frey As I write in the travis.yml as comment, this *android* is deprecated. It looks like, to create a new Android Virtual Device it is only possible to use until android API level 22. Google change a lot in the SDK. For the AVD (Android virtual device) there is also a new command. Try: 

``
/usr/local/android-sdk/tools/bin/avdmanager create avd -n test -k "system-images;android-23;google_apis;armeabi-v7a"
``

See also [avdmanager](https://developer.android.com/studio/command-line/avdmanager.html)

PR #76, opened 2017-11-20
Bugfix for doubles example

quadfault:
The doubles example (sum.cpp, doubles.hpp) wasn't recognizing hexadecimal numbers. A simple swap of the binary and decimal rules seems to do the trick. Also, I changed the name of the binary rule to hexadecimal, since that's what it actually matches.


ColinH:
Thank you!

PR #77, opened 2017-12-04
Fix tracer to work with boolean apply/apply0.

joelfrederico:
The tracer previously only worked with the void versions of apply and apply0. This extends the tracer to be compatible with boolean versions of apply and apply0. Previously, using boolean apply and apply0 resulted in compiler errors when including the tracer when parsing.


d-frey:
Thanks for catching this, just in time for our next release :)

PR #81, opened 2017-12-13
Add missing inline to avoid linking error.

lambdafu:
If I `#include <tao/pegtl.hpp>` in more than one source code file and link them together, I get this error:

```
../lib/liblibneopg.a(http.cpp.o): In function `std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_data() const':
/home/marcus/projects/neopg/neopg/3rdparty/pegtl/include/tao/pegtl/internal/file_reader.hpp:22: multiple definition of `tao::pegtl::internal::file_open(char const*)'
../lib/liblibneopg.a(uri.cpp.o):/home/marcus/projects/neopg/neopg/3rdparty/pegtl/include/tao/pegtl/internal/file_reader.hpp:22: first defined here
clang-4.0: error: linker command failed with exit code 1 (use -v to see invocation)
```

Adding the static helps. Is this the right fix for you?

### DIFF
--- a/include/tao/pegtl/internal/file_reader.hpp
+++ b/include/tao/pegtl/internal/file_reader.hpp
@@ -18,7 +18,7 @@ namespace tao
    {
       namespace internal
       {
-         std::FILE* file_open( const char* filename )
+         inline std::FILE* file_open( const char* filename )
          {
             errno = 0;
 #if defined( _MSC_VER )


d-frey:
I think the correct fix is adding `inline`, not `static`. Also, we don't use tabs, we use spaces. Could you change it accordingly, please?

The real problem is that we don't have a test for this in the CI, though. I guess I should work on that as well, as it is not the first time that we have this kind of problem.

lambdafu:
Ah, sorry, man.  inline doesn't prevent this error, because it does not affect linkage.  In fact, the compiler is forced to generate the symbol anyway, because it does not know if another compilation unit will reference it. You can use static inline, but the inline keyword is pretty much useless nowadays, as compilers do whatever the heck they want (righly so, of course - although gcc has always_inline to force inline).

I knew that, but to my defense, I am currently suffering from a cold. :sneezing_face: Sorry!

lambdafu:
(I deleted a mistaken comment, sorry. I do suffer from a cold though ;)

d-frey:
No problem, but I assume the problem is now fixed for you. If not, feel free to re-open.

lambdafu:
I tested it, and it works for me.  However, I didn't write any code that utilizes the function, so I am not sure what the coverage of my simple test is.

d-frey:
It should be fine now, it's a known problem and a known fix for this kind of problem. See point 1) on cppreference's description of [`inline`](http://en.cppreference.com/w/cpp/language/inline).

PR #85, opened 2018-01-09
Fix broken link "Changing States"

joelfrederico:
Found another broken link for "Changing States", documentation out of sync with index.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -27,7 +27,7 @@
   * [States](Actions-and-States.md#states)
   * [Action Specialisation](Actions-and-States.md#action-specialisation)
   * [Changing Actions](Actions-and-States.md#changing-actions)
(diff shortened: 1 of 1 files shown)


d-frey:
Thanks, again!

Isn't there a way to automate this? On GitHub probably? I'm not specifically asking you, just wondering in general. This seems like such a general problem that GitHub should have a way to figure it out and notify me when I'm committing broken relative links into my own repo...

joelfrederico:
I don't personally know... it seems like there's a few broken links around, a tool would be nice. Somebody has to have run into this problem before.

joelfrederico:
This works: https://github.com/tcort/markdown-link-check

I'll be submitting a PR in a bit with some fixes based on it.

PR #86, opened 2018-01-09
Fix broken links

joelfrederico:
I checked with https://github.com/tcort/markdown-link-check. This is the sort of thing you probably could add to a CI check if you'd like. This should fix all of the currently broken links, minus a gentoo package link that just seems to be a server issue on their end right now.


d-frey:
Thanks, I'll have a look into adding it to the CI later.

joelfrederico:
You're welcome. Sorry I didn't just do this instead of create multiple PR's.

d-frey:
Don't worry about the multiple issues, any help is appreciated. 👍

PR #91, opened 2018-02-14
fix parse_tree node id

michael-brade:
parse_tree: set the node id in `start()`, not in `success()`, otherwise we overwrite the node changes possibly made in `apply()`.


d-frey:
Thanks!

d-frey:
For the default node, I just moved it back to success() as we are no longer using the `apply()` method for parse tree transformations and the `transform` method is called at a different time, so the problem should be gone anyways.

If you still think it is broken, please re-open this issue and provide a testcase against HEAD to demonstrate the problem. Thanks :)

michael-brade:
Nope, you are right! It works :) My fix and your move to `transform()` just overlapped.

PR #110, opened 2018-05-31
Add CMake config file

pleroux0:
Add a pegtl-config.cmake file (and install)

This is required for having a project dependant on PEGTL decoupled from it's conan package.

The project that has PEGTL as a dependancy can use

```cmake
find_package(PEGTL REQUIRED)
target_include_directories(test PUBLIC ${PEGTL_INCLUDE_DIRS})
```

When using system installed PEGTL it will work as expected. Using the conan_find_package generator CMake can find the version of PEGTL from conan.

Edited: Put the cmake code in a code block


d-frey:
Thanks. I have a couple of questions:

* Why is there `PEGTL_INCLUDE_DIR` and `PEGTL_INCLUDE_DIRS`? Wouldn't one be enough? Or is this because of our main `CMakeLists.txt` using `PEGTL_INCLUDE_DIR`, but we missed some (de-facto) standard where everybody else is using `PEGTL_INCLUDE_DIRS` and we should just change it?
* It's yet another file that contains the version number. Could the version number come from the main `CMakeLists.txt` or vice versa? (It's not too big of a deal as we have a release-checklist where we could add just one more point)
* You mentioned Conan: Could we include Conan-support in our main repository directly? Or does it work as-is? If we could add "official" Conan support to the PEGTL, that would be great :)

d-frey:
Actually, if the current way of supporting Conan that you created is the "official" way, that is fine with us. Just let us know how to support you/Conan best. Thanks!

pleroux0:
Feedback:

- From what I understand INCLUDE_DIR is the library itself and INCLUDE_DIRS is the library and it's dependencies. Which in the case of PEGTL is the same. Although trying to find where I first read that made me realize my cmake practices might be slightly outdated. There seems to be nicer ways than textually including required directories. 
- It should be possible to generate the config file with the version string in it. Removing the need for another version string.
- From a package consumer point of view there is no real difference between the package being done out-of-source and in-source. The benefit of having the package done in-source is that the conan package is kept up to date and maintained by the maintainer of the library. This prevents the conan package version from slipping behind. According to [bincrafters](https://bincrafters.github.io/2018/02/27/Updated-Conan-Package-Flow-1.1/) it is a long term goal of the conan project for the packages to be done in-source. There is nothing wrong with it being done out-of-source and many packages are done out-of-source without any issues.

I'll try and remove the need for another version string and brush up on modern CMake practices then update the pull request.

pleroux0:
After a few youtube videos on modern CMake[[1](https://www.youtube.com/watch?v=bsXLMQ6WgIk)][[2](https://www.youtube.com/watch?v=eC9-iRN2b04)]. I've realized the flaws in my old ways and updated the pull-request accordingly.

pegtl is now includable from a cmake project with

```cmake
find_package(PEGTL REQUIRED)
target_link_libraries(test taocpp::pegtl)
```

I didn't make a corrosponding config-version.cmake since I don't know how to do it properly.

d-frey:
Thank you, this is definitely going in the right direction. Don't worry about the version thingy, I can live with it as is.

I will probably watch the CMake-talks you linked tomorrow or at the weekend (running out of time for today) and will probably merge the PR soon.

One question I have right now: Why `target_link_libraries(test taocpp::pegtl)` and not just `target_link_libraries(taocpp::pegtl)`? Or was this just a typo/remnant from your tests?

About doing packaging in-source: I saw that you saw us rejecting some other package manager. The situation is similar to build-systems, we can not support all, but we picked CMake when it was mature and wide-spread enough. And while we can not support all package managers, I feel like Conan is becoming (or already is?) the de-facto standard. We'd be, of course, happy to receive a PR to get us started :)

pleroux0:
I have absolutely no idea why the android-19 test case failed. I don't think it was me.

The example was slightly more minimal than it should've been.

```cmake
cmake_minimum_required(VERSION 3.2)

find_package(PEGTL CONFIG REQUIRED)

add_executable(test test.cpp)
target_link_libraries(test taocpp::pegtl)
```

It automatically makes  the test target compile with the required flags (the includes and setting the standard to c++11).

I think conan is the best package manager out there. I'm not sure if it's considered the de facto standard yet, but I'd be happy to make a pull request making the conan package in-source. Do you want another pull request for that or should I add it to this one?

d-frey:
I've seen Android builds fail before - for no apparent reason. Just ignore it.

You can add to this PR, I'll merge it and do some other cleanups later on as I've seen the first talk you linked and I realize there is more room for improvement.

pleroux0:
To create and upload conan package (Using my remote as an example)

```bash
conan create . pegtl/2.5.2@pleroux0/stable -r=pleroux0
conan upload pegtl/2.5.2@pleroux0/stable -r=pleroux0
```

I used cmake install instead of copying the files to ensure that the pegtl-config.cmake file is installed in the package. I've also left the author field blank as I wasn't sure what exactly you wanted there. I think the taocpp email address is the likely candidate.

d-frey:
Thank you very much!

Would you like to give us your real name for our "Thank You" section or would you prefer to be listed as "pleroux0"?

Issue #111, opened 2018-06-06
Support Conan


uilianries:
**What?** 
Provide PEGTL support to Conan package manager 

**Why?**  
To any user consume PEGTL as Conan binary package.

**How?**  
- Write conanfile.py (recipe)
- Write test_package (recipe validation)
- Run under Travis and Appveyor
- Push to Bintray

uilianries:
Closed by #112

PR #112, opened 2018-06-07
Support Conan

uilianries:
Related issue: #111 

I dont have permission to create branches on taocpp, because I have member role. But no problem, I just forked the project.

This PR just update the support to use Conan + PEGTL

I added a job for Travis and Appeveyor to validate on both platforms

Why I copied the files instead to use cmake helper?  
The cmake helper requires settings.compiler when run over Windows, but this package is header-only and no settings is required. In the future I'll provide FindPEGTL.cmake file, so don't worry.

The next step is update Travis settings. I don't have permission to change the project settings on Travis :( But you need to add `CONAN_LOGIN_USERNAME` as taocpp-user, and `CONAN_PASSWORD` with Key API provided by Bintray.

Regards!

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,9 +1,5 @@
 version: '{branch}-{build}'
 
-branches:
-  only:
-    - master
-
 os:
 - Visual Studio 2015
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -34,9 +34,19 @@ environment:
       COMPILER: MinGW 6
       BINDIR: C:\mingw-w64\x86_64-6.3.0-posix-seh-rt_v5-rev1\mingw64\bin
 
+    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+      GENERATOR: Visual Studio 15 2017
+      CONAN_VISUAL_VERSIONS: 15
+      PYTHON: "C:\\Python27"
+      PYTHON_VERSION: "2.7.8"
+      PYTHON_ARCH: "32"
+      CONAN: TRUE
+
 init: []
 
-install: []
+install:
+- if defined CONAN (set PATH=%PATH%;%PYTHON%/Scripts/)
+- if defined CONAN (pip.exe install -U conan conan-package-tools)
 
 before_build:
 - if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2015" (del "C:\Program Files (x86)\MSBuild\14.0\Microsoft.Common.targets\ImportAfter\Xamarin.Common.targets")
@@ -50,3 +60,4 @@ build_script:
 
 test_script:
 - ctest -C "%CONFIGURATION%" --output-on-failure
+- if defined CONAN (python build.py)
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -45,8 +45,8 @@ environment:
 init: []
 
 install:
-- if defined CONAN (set PATH=%PATH%;%PYTHON%/Scripts/)
 - if defined CONAN (pip.exe install -U conan conan-package-tools)
+- if defined CONAN (conan profile new default --detect)
 
 before_build:
 - if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2015" (del "C:\Program Files (x86)\MSBuild\14.0\Microsoft.Common.targets\ImportAfter\Xamarin.Common.targets")
@@ -60,4 +60,5 @@ build_script:
 
 test_script:
 - ctest -C "%CONFIGURATION%" --output-on-failure
+- if defined CONAN (cd ..)
 - if defined CONAN (python build.py)
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,5 +1,9 @@
 version: '{branch}-{build}'
 
+branches:
+  only:
+    - master
+
 os:
 - Visual Studio 2015
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -425,6 +425,18 @@ matrix:
         - "sed -i 's#TAO_PEGTL_NAMESPACE#pegtl#g' $(find . -name '*.[hc]pp')"
         - make -kj3 clang-tidy
 
+    - language: python
+      python:
+        - "3.6"
+      sudo: required
+      install:
+        - pip install conan conan-package-tools
+      env:
+        - CONAN_GCC_VERSIONS=7
+        - CONAN_DOCKER_IMAGE=lasote/conangcc7
+      script:
+        - python build.py
+
 script:
   - $CXX --version
   - make -kj3
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,6 @@ language: generic
 os: linux
 dist: trusty
 
-branches:
-  only:
-    - master
-
 matrix:
   include:
     - compiler: gcc
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ language: generic
 os: linux
 dist: trusty
 
+branches:
+  only:
+    - master
+
 matrix:
   include:
     - compiler: gcc
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,12 +2,6 @@ cmake_minimum_required (VERSION 3.2.0 FATAL_ERROR)
 
 project (pegtl VERSION 2.5.2 LANGUAGES CXX)
 
-# Build project by Conan
-if(EXISTS "${CMAKE_BINARY_DIR}/conanbuildinfo.cmake")
-    include("${CMAKE_BINARY_DIR}/conanbuildinfo.cmake")
-    conan_basic_setup()
-endif(EXISTS "${CMAKE_BINARY_DIR}/conanbuildinfo.cmake")
-
 # installation directories
 set (PEGTL_INSTALL_INCLUDE_DIR "include" CACHE STRING "The installation include directory")
 set (PEGTL_INSTALL_DOC_DIR "share/doc/tao/pegtl" CACHE STRING "The installation doc directory")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,12 @@ cmake_minimum_required (VERSION 3.2.0 FATAL_ERROR)
 
 project (pegtl VERSION 2.5.2 LANGUAGES CXX)
 
+# Build project by Conan
+if(EXISTS "${CMAKE_BINARY_DIR}/conanbuildinfo.cmake")
+    include("${CMAKE_BINARY_DIR}/conanbuildinfo.cmake")
+    conan_basic_setup()
+endif(EXISTS "${CMAKE_BINARY_DIR}/conanbuildinfo.cmake")
+
 # installation directories
 set (PEGTL_INSTALL_INCLUDE_DIR "include" CACHE STRING "The installation include directory")
 set (PEGTL_INSTALL_DOC_DIR "share/doc/tao/pegtl" CACHE STRING "The installation doc directory")
--- a/build.py
+++ b/build.py
@@ -1,0 +1,29 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import os
+from cpt.packager import ConanMultiPackager
+
+
+class BuilderSettings(object):
+    @property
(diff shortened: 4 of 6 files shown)


pleroux0:
Hi.

Disclosure: I made the conan-pegtl package that was included in bincrafters.

The CMake install generates the target information which is required for use with the cmake_paths generator. Then in CMake:

```cmake
find_package(pegtl)
target_link_library(my_target taocpp::pegtl)
```

Without having to make your CMakeFile.txt dependant on conan or even having to manually set your target to C++11 as CMake knowns that you require it from linking the target to pegtl.

For example if you have package A dependant on pegtl, but package A doesn't use conan then using conan_paths one can make an external conan package for A by just setting the toolchain file when configuring A. This decouples A's build system and conan. It also allows package A to be built without conan.

Don't get me wrong, conan is great, but forcing people to create their libraries specifically as a conan package is not the right way to go.

When building the test package a C++ compiler is necessary so why is it a problem to require a compiler during package()? CMake needs to generate the targets file and for that you need to actually call CMake and not just include the headers. Or is there an alternative to extract the target information from CMake?

In the original conan-pegtl package I had it the way you have it now, but have since discovered that for modern CMake one requires the target file as generated by CMake.

uilianries:
I agree with your idea. Generate the config file makes the package transparent for any usage.

The problem with cmake helper is that it needs settings, so we need to distribute different packages by compiler, but doesn't make sense, because the package is header-only.

I'll ask for help to Conan team, they should know how treat cmake helper in such case.

Regards.

d-frey:
Some remarks on the current patch:

* Would it be possible to drop the `sudo: required` for TravisCI by using `pip install --user ...`?
* For ` test_package/CMakeLists.txt`: Is setting the C++ standard required? IIUC, the dependency should already require it via the features used by `taocpp::pegtl`.
* I agree with @pleroux0 that Conan-support should be optional/transparent for our users. Especially a header-only library should be simple to install. I'm still not sure how well Conan supports header-only libraries, but I hope that for someone using Conan, the process should be as simple as:

1) `conan install pegtl/latest@taocpp/stable` (note: ideally no `--build`)
2) If necessary, explain your system/enviroment to make it visible
3) `#include <tao/pegtl.hpp>` works.

If you are using CMake, it should be as simple as

```cmake
find_package(pegtl)
target_link_libraries(hello taocpp::pegtl)
```

for step 2) (as Paul wrote).

To be honest, I've never managed to make anything work :) If you could help me to understand the missing steps I'd appreciate it.

uilianries:
> Would it be possible to drop the sudo: required for TravisCI by using pip install --user ...?
No, it's not possible for that job. Because it invokes Docker and such action requires root level. But the sudo requirement is to run Docker. The build process, including call Conan, gcc, upload and all things are non-root. They run in the container by Conan user that is non-root level.

> For test_package/CMakeLists.txt: Is setting the C++ standard required? IIUC, the dependency should already require it via the features used by taocpp::pegtl

Good point. I'll remove it.

> I agree with @pleroux0 that Conan-support should be optional/transparent for our users. Especially a header-only library should be simple to install. I'm still not sure how well Conan supports header-only libraries, but I hope that for someone using Conan, the process should be as simple as:

I totally agree with both you. The user should be able to use directly `taocpp:pegtl`. I'll update to use cmake helper for all platforms and keep the same package id.

I just updated the PR as **WIP** to avoid any merge. Please, wait until I finish the new changes. I'll work on the new changes at evening (8h from now).

pleroux0:
To avoid user having to call ``--build pegtl``, the "built" package simply needs to be exported when calling upload. Again using my remote as an example

``bash
conan upload --all pegtl/2.5.2@pleroux0/stable -r=pleroux0
``

would upload the package along with the recipe. Since it is header only there is only one possible package in the cache so we can use ``--all``. I forgot to add this in my previous upload example.

Also ``conan install pegtl/2.5.2@pleroux0@stable`` would only install it into the local cache. It is also not necessary. To use the conan package a conanfile is required. For my use case where I want to be able to build with only CMake and also with conan I could use the following conanfile.txt

```
[requires]
pegtl/2.5.2@pleroux0/stable

[generators]
cmake_paths
```

Then simply with

```bash
conan install ..
cmake -DCMAKE_TOOLCHAIN_FILE=conan_paths.cmake ..
```

You have a CMake project that uses the conan package of pegtl when using ``find_package`` and ``target_link_library``.  After which the ``#include <tao/pegtl.hpp>`` will be that from the conan package.

Importantly **the package works on windows** since it doesn't require building when the package (not just the recipe) is uploaded. It just doesn't want to be created on Windows. Which means this issue doesn't affect consumers of the package. It is something I missed when testing the new CMake install way of doing things method. I'm following conan-io/conan#3011 to see how to resolve that.

pleroux0:
A few comments and questions on your pull request @uilianries 

1. the ``cmake.patch_config_paths()`` does nothing as there isn't an absolute reference to the install directory (intentional). If it had done something it would also have broken transparent use since

```cmake
include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
conan_basic_setup()
```
  would be required in the consumer CMakeLists to set the CONAN variable.  Which isn't always possible. This requirement would break my use case. It is also an experimental feature according to the [documentation](https://docs.conan.io/en/latest/reference/build_helpers/cmake.html?highlight=patch_config_paths#patch-config-paths-experimental).

2. Why is modifying the CMakeLists.txt of PEGTL neccesary? Is it because of  ``cmake.patch_config_paths()``?

3. Is it really necessary to specify the version in the conanfile.py? There are packages (like [ranges-v3](https://github.com/ericniebler/range-v3)) that are great and doesn't have a version specified. I can understand the need for it in out-of-source packages, but since this package is in-source. It's becomes just another version number to keep track of. 

4. Wouldn't it be better to wait for conan-io/conan#3011 to resolve when conan-io/conan#3012 is merged rather than in the mean time adding settings that are unused with no clear indication to the reader why they are there?

5. (More a general question) What are the pros and cons  of auto-uploading whenever there a version change vs. just manually uploading when the maintainer thinks the package is ready for release?

uilianries:
> 1. the cmake.patch_config_paths() does nothing as there isn't an absolute reference to the install directory (intentional). If it had done something it would also have broken transparent use since
 include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
 conan_basic_setup()
 would be required in the consumer CMakeLists to set the CONAN variable. Which isn't always possible. This requirement would break my use case. It is also an experimental feature according to the documentation.

The file `pegtl-config.cmake` generated by the project, uses _IMPORT_PREFIX to solve the header dir, however it's not dynamic. The same file is in the package and should be used by any platform. However, the path is totally static and will not when anyone try to install the package and use taocpp::pegtl target.

The function cmake.patch_config_paths() replaces the hardcoded path in pegtl-config.cmake to use ${CONAN_PEGTL_ROOT}. 

> 2. Why is modifying the CMakeLists.txt of PEGTL neccesary? Is it because of cmake.patch_config_paths()?

conanbuildinfo.cmake is responsable to consume some variables passed by conan:

    CONAN_COMPILER
    CONAN_COMPILER_VERSION
    CONAN_CXX_FLAGS
    CONAN_C_FLAGS
    CONAN_EXPORTED
    CONAN_LIBCXX
    CONAN_SHARED_LINKER_FLAGS

But looking again, they are not required in this case, because the project is header only. I'll revert the cmake file to original state.

> 3. Is it really necessary to specify the version in the conanfile.py? There are packages (like ranges-v3) that are great and doesn't have a version specified. I can understand the need for it in out-of-source packages, but since this package is in-source. It's becomes just another version number to keep track of.

It's not require, but you will need to specify the full reference when creating the package:

    conan create . pegtl/2.5.2@taocpp/stable

Some projects, as Catch2, uses a script update all versions, including in conan and cmake files. I see no problem in use the recipe without a version.

> 4. Wouldn't it be better to wait for conan-io/conan#3011 to resolve when conan-io/conan#3012 is merged rather than in the mean time adding settings that are unused with no clear indication to the reader why they are there?

Yes, this would be great.

> 5. (More a general question) What are the pros and cons of auto-uploading whenever there a version change vs. just manually uploading when the maintainer thinks the package is ready for release?

IMO the package could be released for every new tag. It's easy use Travis for run such filter. I see no problem to upload manually, but is one more task to be executed.

pleroux0:
> The file pegtl-config.cmake generated by the project, uses _IMPORT_PREFIX to solve the header dir, however it's not dynamic. The same file is in the package and should be used by any platform. However, the path is totally static and will not when anyone try to install the package and use taocpp::pegtl target.
> 
> The function cmake.patch_config_paths() replaces the hardcoded path in pegtl-config.cmake to use ${CONAN_PEGTL_ROOT}.

CMake uses relative paths for __IMPORT_PREFIX if a relative path is specified. The generated package works on windows (I tested for exactly for this) as the relative path stays correct. You just set the abs path when using

``cmake.definitions["PEGTL_INSTALL_CMAKE_DIR"] = self.package_folder``

which causes CMake to use an absolute path since your giving it an absolute path.

Only when setting an absolute path (like in your branch) then ``cmake.patch_config_paths()`` adds the CONAN variable. I want to avoid that since it requires coupling any consumer package to conan which isn't neccesary.

I am on cpplang as of a few moments ago.

d-frey:
I've added `CONAN_LOGIN_USERNAME` and `CONAN_PASSWORD`. I guess at some point I should switch the Travis and AppVeyor builds to `taocpp-user` instead of my own account?

uilianries:
@pleroux0 I just remove all stuff that change the cmake cofig file, also the version wont be present on the recipe, and only tag will be uploaded.

You are right about PEGTL_INSTALL_CMAKE_DIR, so I just removed from the patch.

As you commented about ranges-v3, I remove the version attribute from the recipe. Although, the same must be solved when the package is created. When running on user machine it could be:

    conan create . regtl/2.5.2@taocpp/stable

As you can see, the reference is full, but I think that is not a problem. During the CI, the script build.py will extract the project version from CMake file.

And finally, to avoid multiple uploads when some new feature is merged on master, the pattern used to upload the package, will accept only tag format, number+.number+.number+. I could apply a simple rule on travis.yml to run the Conan job only when a tag is created, but IMO I prefer to run for every build to make sure about any regression.

uilianries:
@d-frey Yes, for now it's okay, because I don't have permission to access the project settings on Travis. But, you should use the taocpp-user soon, to avoid to expose your profile information.

Also, I don't have permission on Appveyor to stop/cancel any build. I would like to have this access, because if I submit multiple changes, I need to wait for each build, even the outdated.

uilianries:
I think that the recipe is good to go.

The issue https://github.com/conan-io/conan/issues/3011 only will be solved on next month, because it target is Conan 1.5.0. So, we could release this recipe and update after to have the new Conan release.

WDYT @pleroux0 @d-frey ?

uilianries:
Hi @pleroux0 !

I just rebased my [branch](https://github.com/uilianries/PEGTL/tree/feature/conan) over [master](https://github.com/taocpp/PEGTL/tree/master) and I have the same result as before, all is working fine.

To reproduce my steps:

    $ docker run --rm -ti lasote/conangcc7
    $ https://github.com/uilianries/PEGTL.git
    $ cd PEGTL
    $ git checkout feature/conan
    $ conan create . pegtl/2.5.2@taocpp/stable

You could check the log [here](https://gist.github.com/uilianries/626da17f6b1ce28dc9e78f978ee41f48)

> Did you perchance rebase and force push onto an existing branch?

Yes, I "squashed" the Appveyor commits. Unfortunately, there is no way to validate Appveyor config on local machine, before to push.

pleroux0:
@uilianries 

The rebase must've confused my local git repo in a new clone it works with ``conan create``. I was additionally confused by ``gitg`` which showed the following

image

after merge (looks like a bug in gitg). ``git log --graph``shows the correct structure after merge. Sorry for the confusion. Everything is good.

uilianries:
@pleroux0 No problem, It was my fault, push --force is ugly IMO, sorry about that

taocpp-user:
The values for `CONAN_LOGIN_USERNAME` and `CONAN_PASSWORD` should be correct.

For AppVeyor, I added the taocpp/maintenance Team from GitHub and gave it Administrator rights. Is that sufficient? (I think this should be the right way, as you are a member of that team)

uilianries:
@d-frey Now I'm able to stop job on Appveyor! Thanks!

PR #116, opened 2018-06-13
Move Conan stuff to custom path

### LINKED ISSUE #114: Move Conan stuff to sub directory
To keep pegtl project clean, we should move build.py and test_package to a new sub-directory: conan

Conan solves the test directory by the names test or test_package, on same dir level, to use a custom path, the option -tf should be used.

Conan package tools has no support by env var to solve it, so we need to use the arguments.

uilianries:
Hi!

This PR solves the issue #114 

After to have Conan Package Tools 0.18.0, we are able to use test_package in a custom path.

Close #114 

Regards!


d-frey:
I now see both TravisCI and AppVeyor running twice!? This delays feedback and waste resources.

Can we do something about this? (I have to go to work now, if you know how to fix it please go ahead, otherwise I'll have a look an the weekend)

uilianries:
> I now see both TravisCI and AppVeyor running twice!? This delays feedback and waste resources.

It's because one job is related for the feature branch, and the second is for the PR.

On my last commit, I used a fork, so the second job didn't appear on PR list.

Issue #120, opened 2018-07-11
Unit test failure on s390x architecture


dkopecek:
Hi,
 I've tried to run the PEGTL unit tests on a Linux s390x machine and got the following errors:
```
pegtl: unit test assert [ internal::h_to_le( std::int16_t( 0x7a39 ) ) == std::int16_t( 0x7a39 ) ] failed in line [ 53 ] file [ src/test/pegtl/internal_endian.cpp ]
pegtl: unit test assert [ internal::h_to_le( std::uint16_t( 0x7a39 ) ) == std::uint16_t( 0x7a39 ) ] failed in line [ 54 ] file [ src/test/pegtl/internal_endian.cpp ]
pegtl: unit test assert [ internal::h_to_be( std::int16_t( 0x7a39 ) ) == std::int16_t( 0x397a ) ] failed in line [ 55 ] file [ src/test/pegtl/internal_endian.cpp ]
pegtl: unit test assert [ internal::h_to_be( std::uint16_t( 0x7a39 ) ) == std::uint16_t( 0x397a ) ] failed in line [ 56 ] file [ src/test/pegtl/internal_endian.cpp ]
pegtl: unit test assert [ internal::le_to_h( std::int16_t( 0x7a39 ) ) == std::int16_t( 0x7a39 ) ] failed in line [ 57 ] file [ src/test/pegtl/internal_endian.cpp ]
pegtl: unit test assert [ internal::le_to_h( std::uint16_t( 0x7a39 ) ) == std::uint16_t( 0x7a39 ) ] failed in line [ 58 ] file [ src/test/pegtl/internal_endian.cpp ]
pegtl: unit test assert [ internal::be_to_h( std::int16_t( 0x7a39 ) ) == std::int16_t( 0x397a ) ] failed in line [ 59 ] file [ src/test/pegtl/internal_endian.cpp ]
pegtl: unit test assert [ internal::be_to_h( std::uint16_t( 0x7a39 ) ) == std::uint16_t( 0x397a ) ] failed in line [ 60 ] file [ src/test/pegtl/internal_endian.cpp ]
pegtl: unit test assert [ internal::h_to_le( std::int32_t( 0x7a391f2b ) ) == std::int32_t( 0x7a391f2b ) ] failed in line [ 62 ] file [ src/test/pegtl/internal_endian.cpp ]
[...]

PR #121, opened 2018-07-11
Fix endianness detection in internal_endian unit test

dkopecek:
Resolves: https://github.com/taocpp/PEGTL/issues/120


ColinH:
Thanks again, guess it shows that we haven't had access to a big-endian machine for a long time.

PR #123, opened 2018-07-24
Add Win32 memory mapped file support

studoot:
I've added the Win32 equivalent of Unix `mmap` support, using the equivalent set of Win32 functions (`CreateFileA`, followed by `CreateFileMappingA` and finally `MapViewOfFile`. I've added a header (`win32_file_mapper.hpp`) which contains RAII classes for the resources opened by these 3 functions (`win32_file_opener` => `CreateFileA`, manages a Win32 `HANDLE`, `win32_file_mapper` => `CreateFileMappingA`, manages a Win32 `HANDLE`, and then `file_mapper` => `MapViewOfFile`, manages the pointer that is the handle for the mapped view). I didn't want to add header files for the 'private' classes (`win32_file_opener` and `win32_file_mapper`), as they're only going to be used by `file_mapper`.

Other than that, the changes are:

1. `file_input.hpp`, `mmap_input.hpp`, `file_mmap.hpp` - enable mmap support if the macro `_WIN32` is defined (as it will be by MSVC & (I'm pretty sure)MINGW32).

2. `input_error.hpp` - add a macro to throw an `input_error` with the result of `GetLastError()` (the Win32 equivalent of `errno`). This is used only within `win32_file_mapper.hpp`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ CXXFLAGS ?= -Wall -Wextra -Wshadow -Werror -O3 $(MINGW_CXXFLAGS)
 
 CLANG_TIDY ?= clang-tidy
 
-HEADERS := $(filter-out include/tao/pegtl/internal/endian_win.hpp,$(shell find include -name '*.hpp')) $(filter-out src/test/pegtl/main.hpp,$(shell find src -name '*.hpp'))
+HEADERS := $(filter-out include/tao/pegtl/internal/endian_win.hpp include/tao/pegtl/internal/win32_file_mapper.hpp,$(shell find include -name '*.hpp')) $(filter-out src/test/pegtl/main.hpp,$(shell find src -name '*.hpp'))
 SOURCES := $(shell find src -name '*.cpp')
 DEPENDS := $(SOURCES:%.cpp=build/%.d)
 BINARIES := $(SOURCES:%.cpp=build/%)
--- a/include/tao/pegtl/file_input.hpp
+++ b/include/tao/pegtl/file_input.hpp
@@ -12,7 +12,7 @@
 #include <unistd.h>  // Required for _POSIX_MAPPED_FILES
 #endif
 
-#if defined( _POSIX_MAPPED_FILES )
+#if defined( _POSIX_MAPPED_FILES ) || defined( _WIN32 )
 #include "mmap_input.hpp"
 #else
 #include "read_input.hpp"
@@ -22,7 +22,7 @@ namespace tao
 {
    namespace TAO_PEGTL_NAMESPACE
    {
-#if defined( _POSIX_MAPPED_FILES )
+#if defined( _POSIX_MAPPED_FILES ) || defined( _WIN32 )
       template< tracking_mode P = tracking_mode::IMMEDIATE, typename Eol = eol::lf_crlf >
       using file_input = mmap_input< P, Eol >;
 #else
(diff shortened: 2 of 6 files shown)


nshcat:
It's a bit sad that this pulls in the WinApi header inclusion into your source files. But I guess there is no way around that given the fact that the PEGTL is header only..

studoot:
>It's a bit sad that this pulls in the WinApi header inclusion into your source files. But I guess there is no way around that given the fact that the PEGTL is header only..

@nshcat The file it’s in is only included if _WIN32 is defined, and that file could be enabled/disabled by a feature macro?

But yes, header only makes it hard to hide Windows.h 😬

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

nshcat:
Yes, the guard is good practice. This PR is actually quite nice because I need mmap support on windows, but I was thinking about how a non-header-only-library could use the pimpl pattern to hide that dependency. Maybe I am just going to create my own wrapped around this of this PR is accepted :)

d-frey:
Thanks for working on this, but it seems the patch needs more work.

Looking at the CI failures, clang-tidy isn't too happy about a few things and the Windows headers need to be explicitly excluded from clang-tidy as it won't be able to find the headers. Look at the Makefile line 37.

Also, you need to protect against already set `NOMINMAX` and only `#undef` it if it wasn't set previously.

It also seems you have broken one testcase in some configurations.

Please fix what you can and let me know if you need help.

studoot:
OK.

1. Fixed the `NOMINMAX`, `WIN32_LEAN_AND_MEAN` and `_X86_` macro definitions.
2. Modified the `Makefile` to exclude `win32_file_mapper.hpp` from clang-tidy checks
3. Changed the `#if` structure I added to `mmap_input.hpp` back to `#include file_mapper.hpp` unless `_WIN32` is defined. I think (hope?!) this'll fix the Android unit test seg fault.

studoot:
1 failure - g++-8/cppcheck failed due to `apt install` failing - looks like a connectivity issue?

```
W: Failed to fetch http://mirror.jmu.edu/pub/ubuntu/dists/trusty/InRelease  Could not connect to mirror.jmu.edu:80 (134.126.36.183), connection timed out
W: Failed to fetch http://mirror.jmu.edu/pub/ubuntu/dists/trusty-updates/InRelease  Unable to connect to mirror.jmu.edu:http:
W: Failed to fetch http://mirror.jmu.edu/pub/ubuntu/dists/trusty-backports/InRelease  Unable to connect to mirror.jmu.edu:http:
W: Failed to fetch http://mirror.jmu.edu/pub/ubuntu/dists/trusty-security/InRelease  Unable to connect to mirror.jmu.edu:http:
W: Some index files failed to download. They have been ignored, or old ones used instead.
travis_fold:end:apt-get.diagnostics
[0K
[31;1mThe command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $TRAVIS_APT_OPTS install g++-8 cppcheck" failed and exited with 100 during .[0m
```

studoot:
Hmmm - well, there's [one android build failing](https://travis-ci.org/taocpp/PEGTL/jobs/407940762) (seg fault in test `rule_star_must`) which doesn't fail when [I pass it through Travis in a different branch in my fork](https://travis-ci.org/studoot/PEGTL/builds/407968892). The [clang-6.0 test with `CPPFLAGS=-fsanitize=undefined`](https://travis-ci.org/taocpp/PEGTL/jobs/407940769) can't `apt install` some stuff (and that test also passed successfully in my branch).

The AppVeyor tests all pass now, though, which is nice...

However, I've had to go back to `#include`ing `Windows.h`, as some of the failing AppVeyor tests (the MinGW ones) were because the Windows SDK they use didn't have the same header layout as the newer ones used by Visual Studio 2015/7. One of those things.

d-frey:
OK, all tests passed. Thanks!

PR #125, opened 2018-08-11
Fix cross-compiling to Windows from Linux with MinGW

pleroux0:


d-frey:
For me to understand: Is this because file names are case-insensitive on Windows but case-sensitive when you are compiling on Linux?

PR #126, opened 2018-08-20
typo fixed in doc

delpinux:
space was referred as equivalent to one< ..., 't', ... > instead of one< ..., '\t', ... >


delpinux:
Looks like travis fails for some non reason not related to my change:
- g++-4.8 build seems stuck as soon as job starts
- clang-6 / c++17 fails in the apt-get process
Not sure what to do about it.

d-frey:
Thanks! And just ignore the CI errors, those spurious failures happen from time to time.

Issue #71, opened 2017-09-06
Need hierarchical action handling


thelema:
I'm writing a parser for the Verilog language using PEGTL and wanted to share a problem I'm running into and a possible solution.  This is my first time using PEG-style parsing, although I am far from a novice at parsing, having done work for my PhD thesis on approximate parsing in a network security context (google "flowsifter").

I can appreciate the simplicity and elegance (and incredible run-time efficiency) of having actions associated with each rule run when that rule is matched; this combines with the control system to make a very powerful way to write code that gets triggered by parsing.  But I'm particularly annoyed at how unnecessarily difficult it is to write side-effect heavy code that works correctly without these pieces of code having a hierarchical context to work in. 

In the best parsers I've worked with (and written), when part of the grammar matches, it can return a value to the next level up, and that value can be used as part of constructing the result of that higher-level rule's parse result.  This ends up resulting in a sequence of function calls that are nested in exactly the same structure as the parse tree of the text being parsed.  In PEGTL, I can decide exactly what code runs when a particular rule matches, but I get no nesting of function calls, I only get a flat space of code executions.  Of course it's possible to implement that hierarchy by using an explicit stack and pushing and popping from the stack as rules are matched, but this has a high degree of complexity and is prone to user error.  The expression parser example in this repository even goes as far as using a stack of stacks to handle parentheses in expressions.  

There's got to be a reasonable way to stitch together rules and actions in such a way that actions can return a value and the action for a rule receives/can access values produced by child rules.  Maybe this will require each child rule to have an action that returns a value; that seems a reasonable price to pay for this feature.

ColinH:
This is an interesting subject, and one that has been on the back of our minds for a while... Without having spent too much time thinking about it, my feeling is that it might not be a good fit for the design and structure of the PEGTL. However this might just be a sign that it needs some more thought, and some serious experiments to find a "nice" way of adding this. Have you already tried something in this direction, or have any ideas what to try? Even just showing how you would like to write things could be helpful.

grandquista:
I am currently solving this with the `change_state` control helper. Each nesting gets a new `std::stack` state with the template type matching the grammar construction. The `success` static performs mutations to get child structures into the parent stack. There are only a few `action` specializations to construct leaf nodes. Global and local failures revert state to the last successful stack, so cleanup is automatic.

This has provided type safety and nested structures in my code.

Also global parsing structures are stored in an `input` wrapper to get around the state constantly changing.

[...]

PR #127, opened 2018-08-30
[WIP] parse_tree apply for hierarchical actions and match node construction

mkrupcale:
### Introduction

This is some work that I've been pursuing in order to allow for the construction of nodes only when there is a successful `match` rather than on every `start` which might later fail to `match`, reducing the number of heap allocations. In this pursuit, I implemented some form of hierarchical actions. However, the design is a bit more complicated and has a state which uses a `std::map` in order to keep track of the hierarchy, mapping a level to a span of nodes in the node stack.

### Summary of changes

The changes are right now divided into 5 separate commits:

1. This stylistic change is for consistency with other parts of the PEGTL code such as `parse.hpp`
2. Adds an Action template template parameter to `parse`
3. Adds `make_action` which works similar to `make_control` and is based on both an Action class and a Node class which may implement any combination of the `apply`/`apply0` methods returning `void` or `bool`
4. Uses the Node `apply` method to assign content rather than the `start` and `success` methods to assign the begin and end iterators, respectively
5. Uses the `make_action` methods to construct nodes only after matching and successful Action `apply`/`apply0`

Commits 1-3 can really stand on their own and essentially create the framework for hierarchical actions which was necessary to construct the nodes on `match`. I'm not sure if this addresses issue #71 completely, but it at least might serve as a good starting point.

### Performance

In some preliminary performance tests I've run, it seems like the performance is often similar but sometimes as much as 20% slower. The current design must do a lookup via `std::map::find()` of the current level and/or level below the current level in several places. I suspect that this can be improved with some optimization, however, by keeping better track of the `std::map` iterator when moving up and down the hierarchy. In any case, I haven't been able to come up with the right optimization so far, so I think it would be useful to get some additional eyes looking at this to see if someone can come up with some good optimizations.

### Conclusions

I would appreciate some feedback on whether this is useful at all, and if so, I would like to see if anyone has ideas on how it can be improved and optimized.

### DIFF
--- a/include/tao/pegtl/contrib/parse_tree.hpp
+++ b/include/tao/pegtl/contrib/parse_tree.hpp
@@ -18,6 +18,8 @@
 
 #include "../internal/conditional.hpp"
 #include "../internal/demangle.hpp"
+#include "../internal/has_apply.hpp"
+#include "../internal/has_apply0.hpp"
 #include "../internal/iterator.hpp"
 
 namespace tao
@@ -106,7 +108,7 @@ namespace tao
                m_begin = in.iterator();
             }
 
-            // if parsing of the rule succeeded, this method is called
+            // if application of the rule succeeded, this method is called
             template< typename Rule, typename Input, typename... States >
             void success( const Input& in, States&&... /*unused*/ ) noexcept
             {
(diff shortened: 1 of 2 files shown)


d-frey:
I manually applies the first two patches, including some further cleanups and renamings, and fixing one little (unrelated) bug.

The third commit of yours is where I'm slightly lost. It seems to add a lot of complexity and I wonder whether having two different levels of `apply`-methods is a good idea. Also taking into account the other patches it seems that you might not need the parse-tree infrastructure that much after all. Could you not just base it on the normal `parse`-method with your own Action class template and States?

Also, the state you created seems to collect the subsequent nodes and either discards them or moves them to a node on success - but moving here is cheap and I don't see how it'll save that many allocations.

Finally, if you do supply an Action class template and States, you now added an additional state-parameter passed which is inconsistent and unintuitive. I would like to see this being avoided.

mkrupcale:
> I manually applied the first two patches, including some further cleanups and renamings, and fixing one little (unrelated) bug.

Okay, sounds good.

> The third commit of yours is where I'm slightly lost. It seems to add a lot of complexity and I wonder whether having two different levels of apply-methods is a good idea.

I agree that this is quite complex, but I felt the flexibility of two different types of `apply` methods (ones external to the Node, and one which is a member function of the Node) might be useful for situtations where you want to have better control over Node creation. In particular, you'll notice that the Node `apply` methods only get run if the external `apply` is successful (returns `true` for the `bool apply` methods). In commit 5 this is actually used to avoid additional allocations (`state::create_parent`) when the external `apply` method fails. Additionally, one may or may not need access to the Node private members when applying the action, so having the possibility of both could be useful in this situation as well.

> Also taking into account the other patches it seems that you might not need the parse-tree infrastructure that much after all. Could you not just base it on the normal parse-method with your own Action class template and States?

I'm not quite sure I understand you here, but of course, the `parse_tree` infrastructure is nothing but what you just described: a state which keeps track of hierarchical information and constructs Nodes as part of the state. Currently, this is done exclusively with the lower-level `make_control::start`, `make_control::success`, and `make_control::failure` methods. The current PR, however, seeks to add to this with hierarchical actions which allow for more control of the Node construction. Based on the PEGTL documentation[1,2], Actions also appear to be the main intended user-customization point of the Rule matching, so I don't see why we should not allow the user to easily do this. Since you've applied a variant of commit 2 here, it is now possible for users to supply such external Actions but not Node Actions.

> Also, the state you created seems to collect the subsequent nodes and either discards them or moves them to a node on success - but moving here is cheap and I don't see how it'll save that many allocations.

Here, the Nodes collect _previous_ (children) Nodes created, not subsequent (in the `stack`) ones as the current implementation does. There are actually two places, then, that allocations are saved:
1. As I mentioned in my introductory post, Nodes are only created on `match` rather than `start`, which means that we know that this Rule actually matched. Previously, every attempted Rule would create a Node, regardless of whether it will successfully `match`. This seems like unnecessary allocations to me, although it does allow you to have the Node `start` and `failure` member functions since you know the Node will exist on `start`.
2. Since we are now creating Nodes only after they `match`, this means (based on how PEGTL does the matches) that we must have successfully matched all of the children Nodes, and therefore we know exactly how many `children` Nodes to allocate for the parent Node creation (`state::create_node` and `state::create_root`). Currently, you have to re-allocate the `std::vector< std::unique_ptr< node_t > > children` everytime you `emplace_back( move(...) )` more children in `make_control::success` than its capacity (which starts at 0 by default for most implementations). Thus, for hierarchies with large numbers of children on each Node (large degree), this can save many re-allocations.

As you point out, the moves are cheap, and that's exactly what we seek to utilize.

> Finally, if you do supply an Action class template and States, you now added an additional state-parameter passed which is inconsistent and unintuitive. I would like to see this being avoided.

I'm not quite sure I understand you here. The `States` parameter is any additional state that the user wishes to maintain outside of the `parse_tree`, same as any other PEGTL parser. I don't think we want to prevent the user from maintaining state outside of the `internal::state` necessary for the `parse_tree` creation.

[1] [Actions and States](https://github.com/taocpp/PEGTL/blob/master/doc/Actions-and-States.md)
[2] [Control and Debug](https://github.com/taocpp/PEGTL/blob/master/doc/Control-and-Debug.md)

d-frey:
I've had a closer look at what your patch does and I do think there are some good ideas in there. However, the initial description is quite misleading (at least to me). Node will be created (as they must be) even when the parent rule will not match. This is easy to see when you consider

```c++
struct E : seq< A, B, C, D > {};
```

Obviously, when `D` fails to match you already did create nodes for `A`, `B` and `C`. Anyways, I see that you are trying to reduce the number of intermediate nodes created and discarded. For that, you have created a quite complex solution and I am wondering about these points:

1) Is allowing `apply` and `apply0` with `void` or `bool` return types necessary for the nodes? Or could we massively reduce the complexity of the patch by simply assuming a `void apply()` method for the nodes? Do we even expect this interface to vary much? I'd probably consider renaming it to `assign`, as I don't see this part of the infrastructure to be similar enough to the Action class template. And there wouldn't be any need here to detect the presence of such a function, it must be there for each Rule.

2) Instead of having the map and the spans, it should be enough to have just one vector/stack for the nodes where they are temporarily stored (like in the above example for `E`) and one vector of iterators where the levels in the other vector/stack start (their indices).

3) With the above simplifications, the code might not just be smaller and easier to understand, it might also be more efficient. From what I understand efficiency is the goal of this patch, if there are other goals, please help me understand those as well.

mkrupcale:
> I've had a closer look at what your patch does and I do think there are some good ideas in there. However, the initial description is quite misleading (at least to me). Node will be created (as they must be) even when the parent rule will not match. This is easy to see when you consider
> ```C++
> struct E : seq< A, B, C, D > {};
> ```
> Obviously, when `D` fails to match you already did create nodes for `A`, `B` and `C`.

Indeed this PR does not try to look ahead to determine if every child (in the case of `seq`) will match or anything like that. This would be essentially like doing the complete parse because you'd have to look ahead to the end of the grammar and match the root, only creating the Nodes once you've finished parsing the entire input. Perhaps this _can_ be done, but if you're creating a parse tree, you still need to somehow keep track of the hierarchical information and potentially store or process the actual match text. Since the hierarchical information is generated in the call stack of the PEGTL grammar parser, you might as well create the Nodes when they match.

> Anyways, I see that you are trying to reduce the number of intermediate nodes created and discarded.

Yes, this does the best we can by saying: "This Rule definitely matched, so without any additional information this far into the parse, I am going to create a Node for this Rule. If the parent Rule fails to match, then that child Node will have to be deleted later."

In contrast, currently, it is saying: "This Rule _might_ match, so I will create a Node for this Rule. If this Rule fails to match, then that Node will have to be deleted."

The latter clearly allocates at least as many Nodes as the former since its requirement for Node creation is looser and the parser is less informed at the point of creation.

> For that, you have created a quite complex solution and I am wondering about these points:
> 1. Is allowing apply and apply0 with void or bool return types necessary for the nodes? Or could we massively reduce the complexity of the patch by simply assuming a void apply() method for the nodes? Do we even expect this interface to vary much? I'd probably consider renaming it to assign, as I don't see this part of the infrastructure to be similar enough to the Action class template. And there wouldn't be any need here to detect the presence of such a function, it must be there for each Rule.

Commit 4 shows an example usage of the `void apply()` Node method in the `basic_node`, which makes sense for the general case as you suggest, taking the ActionInput and setting the begin and end iterators for the matched text. Some Rules, however, may not need to know the matched text and therefore might not need ActionInput passed as a parameter and can use `apply0`. An example of this might be a Rule

```C++
struct plus : pegtl::pad< pegtl::one< '+' >, pegtl::space > {};
```

for a calculator for which the padding and the '+' (or any additional bytes for other character encodings) are irrelevant for the interpreter. This might allow for some optimizations compared to `apply()`[1]. Similarly, this Rule may not even need any Actions, but we still need to construct a Node for this Rule, so we still want to check for the presence of `apply0` or `apply`, similar to external Actions.

Whether it's necessary for a Node Action to potentially fail (and thus return `bool`) I suppose depends on the complexity of that Node Action. Suppose that the Action actually is some form of interpreter which evaluates whether or not the current match makes sense in the given context (i.e. this is not built into the grammar itself since it would unnecessarily complicate the grammar for some edge case). A concrete example of this is in the Python grammar[2]: their `atom_expr` has several possible `atom`s each with zero or more `trailer`s. Thus, `0()` is valid Python grammar, but it is invalid as far as the interpreter is concerned: `TypeError: 'int' object is not callable`.

In this case it might make sense for the Node Action to fail and stop the parser from continuing. Furthermore, this may need to be a Node Action and not an external Action because you would like to have access to the children of that Node for its interpretation.

> 2. Instead of having the map and the spans, it should be enough to have just one vector/stack for the nodes where they are temporarily stored (like in the above example for `E`) and one vector of iterators where the levels in the other vector/stack start (their indices).

If I understand you correctly here, you're saying that the existing stack (as per this PR) would still store the Nodes as they are created, but a second vector (instead of the map) would store the spans at a given level. Thus, the index of this second vector is the level in the hierarchy, yes?

I believe I had considered this, and it has the nice advantage of O(1) lookups and inserts (once you've allocated the vector of sufficient size), resolving the performance problem of keeping track of the iterator to the level map I mentioned in my first post. However, this has the disadvantage of possibly requiring more memory. Consider first the `std::map` and second the `std::vector`. Then the memory required by the hierarchy storage is:

1. `sizeof(std::map) + (sizeof(std::size_t) + sizeof(span) + sizeof(std::_Rb_tree_node_base)) * num_active_levels = 48 bytes + (8 bytes + 16 bytes + 32 bytes) * num_active_levels = 48 bytes + 56 bytes * num_active_levels`
2. `sizeof(std::vector) + sizeof(span) * max_depth_of_grammar_match = 24 bytes + 16 bytes * max_depth_of_grammar_match`

where `num_active_levels` is the number of levels currently stored (note that this PR has levels deleted when they are no longer used: after collecting, elevating, or erasing children level or the current node level with no siblings), and `max_depth_of_grammar_match` is the maximum depth of any Node which matches in the grammar. For a recursive grammar, this depth could be large, whereas the number of active levels is usually small, unless the text being matched has a high asymmetry: imagine "1+(2+(3+(4+(...))))" where we can't collect children until we have gone down the right branches all the way to another leaf (i.e. not an enclosure).

I'd certainly be open to trying this approach, though, if we can figure out how to ensure that we don't try to allocate insane amounts of memory for deep matches.

> 3. With the above simplifications, the code might not just be smaller and easier to understand, it might also be more efficient. From what I understand efficiency is the goal of this patch, if there are other goals, please help me understand those as well.

Efficiency is certainly one goal of this patch but hierarchical actions is another. In particular, my above examples might shed light on some interesting things one could do with hierarchical Node Actions: instead of having the parser generate the parse tree or AST and then have an interpreter traverse the completed tree, the traversal could occur during the parse since sometimes (often?) the interpreter only needs the children of a given node. I'm not sure if this would yield performance improvements versus two separate "traversals" (one for parse construction and one for interpretation), but I thought it might be interesting to try.

Reducing the code size and complexity might indeed yield performance improvements, but this design leaves that to the user: they will only get the level of complexity that they request from their Node Action and external Actions through `make_action`. They could have no external Action (e.g. `nothing`) and then have only an `apply0` for some Rules and `apply` for others that need the matched text.

[1] [Actions and States - Apply0](https://github.com/taocpp/PEGTL/blob/master/doc/Actions-and-States.md#apply0)
[2] [10. Full Grammar specification &#8212; Python 3.7.0 documentation](https://docs.python.org/3/reference/grammar.html)

d-frey:
Lets start with point 3: IIUC you are looking for a way to avoid the two-pass approach that usually comes with a parse tree. But that will most likely not work in the general case: The user will provide states `st...` that might be modified when an `apply`-method is called. If the node will later be deleted because a higher level rule did not match after all, these modification will not be reversed. Hence the user must still track the rewinding and reverse the state manually.

It also means that you might get closer to hierarchical actions, but not fully achieve it, at least not in the way I understand the goal of hierarchical actions: Hierarchical actions will be applied to the final parse tree (or AST), there should be no doubt about whether any node will be deleted later on. The final parse tree is only known after the full parse run, hence it must be a two-pass approach to apply actions on it (and only it).

Which leaves us with efficiency, but it seems the current's patch performance is no better than the original code. Still the original code generated a lot of unnecessary nodes and it makes sense to think about ways to get rid of them. I'll see when I find more time for that.

mkrupcale:
> Lets start with point 3: IIUC you are looking for a way to avoid the two-pass approach that usually comes with a parse tree. But that will most likely not work in the general case: The user will provide states st... that might be modified when an apply-method is called. If the node will later be deleted because a higher level rule did not match after all, these modification will not be reversed. Hence the user must still track the rewinding and reverse the state manually.

Yes, for the reason you described I don't think this will work for the general case using an external State `st...`. However, if the user instead creates a custom Node class which stores in private member variables whatever state it needs, then the Node Action can use those internal variables. Any Node state modifications made by `apply` will then be reversed when the Node gets deleted if the parent Rule fails to `match`. So the user can use States `st...` only in the external Action I suppose.

Depending on the complexity of the Node Action, this might be a lot of work to waste in the case that it does get deleted eventually, though. So perhaps it would be better just to wait until the final parse tree or AST to do the full interpretation. However, it still might make sense to do some simple interpretation, depending on the Rule, do decide whether the Node `match` is valid in the given context (like the Python example above).

> It also means that you might get closer to hierarchical actions, but not fully achieve it, at least not in the way I understand the goal of hierarchical actions: Hierarchical actions will be applied to the final parse tree (or AST), there should be no doubt about whether any node will be deleted later on. The final parse tree is only known after the full parse run, hence it must be a two-pass approach to apply actions on it (and only it).

Perhaps yes the final interpretation must be done on the completed parse tree or AST, but that does not mean that hierarchical actions as currently implemented by this PR do not achieve something useful. If the Node is to be created on `match` rather than `start`, then the Node text content must be assigned in `apply` as in Commits 4-5. And as I indicated above, maybe additional Node processing during the parse might be useful to terminate the parse if there is a problem not with the grammar but with something else.

> Which leaves us with efficiency, but it seems the current's patch performance is no better than the original code. Still the original code generated a lot of unnecessary nodes and it makes sense to think about ways to get rid of them.

I wouldn't quite say that the current PR performance is no better than the original code. I think it's fairer to say it's better in some cases and worse in other cases (at least in my early testing). Given that this PR reduces the number of Node allocations, I think it's a matter of finding the right optimization of the state and efficiently keeping track of hierarchical information to yield a universal performance gain. Perhaps this means not using a `std::map` or just keeping better track of its iterators as part of the internal state, but I think this PR serves as a good starting point.

> I'll see when I find more time for that.

Okay, sounds good. I'm looking for new ideas on how to improve this.

d-frey:
I just committed a completely different approach to reducing the number of intermediate nodes.

When I was measuring the number of nodes, I found that our `abnf2pegtl.cpp` example (used with `abnf.abnf`) was originally creating ~24,000 nodes (for a final parse tree with 300 nodes).

Your PR reduces that to ~10,000 nodes but at quite some cost.

My new approach initially used only the `is_final_node` (see the current code in master) which also reduced the number of nodes to ~10,000. But it is much simpler and works at compile-time, there is not run-time overhead at all.

I then added another level (`is_final_node_2`, yes, stupid name) and that reduces the number to ~6,500 nodes. Again, all is done at compile-time.

Since your PR also provide additional features, you might want to consider whether some of my optimizations can be merged into it. Let me know what you think and how we should continue.

mkrupcale:
I'll start out by discussing my understanding of your optimizations to make sure my understanding is correct.

## Understanding of optimizations

### fbf951d09fe58731f5646b1b6892f8d9685bc51a
`is_final_node` essentially selects rules which are "final" such as simple rules like `one`, `ranges`, or `bytes`, and when these rules are not selected, since they are "final" in the grammar (they can't have any children). In the terminology of a tree data structure, "final" might be understood as "leaf", yes? Of course, whether or not the Rule's Node ends up being an actual leaf in the tree will depend on any transforms made and such.

This optimization of not allocating Nodes, keeping track of the `stack`, and moving children around etc. only helps in the case that you are not storing the final node (i.e. `Selector< Rule >::value = false, is_final_node< typename Rule::analyze_t >::value = true`), though, yes?

### 9782e91d6355064dc15df6216c50ec24622530d5
A minor suggestion:
```C++
template< bool... Bs >
using is_all = std::is_same< bool_sequence< Bs..., true >, bool_sequence< true, Bs... > >;
```
This adds an unnecessary extra bool template parameter: you could instead use
```C++
template< bool... Bs >
using is_all = std::is_same< bool_sequence< Bs... >, bool_sequence< ( Bs || true )... > >;
```

If I understand this optimization correctly, the reason that this will further reduce the number of nodes allocated is that compared to the templates

```C++
template< analysis::rule_type Type >
struct is_final_node< analysis::generic< Type > >
   : std::true_type {};

template< analysis::rule_type Type, unsigned Count >
struct is_final_node< analysis::counted< Type, Count > >
   : std::true_type {};
```

the templates

```C++
template< analysis::rule_type Type, typename... Rules, template< typename... > class Selector >
struct is_final_node_2< analysis::generic< Type, Rules... >, Selector >
   : is_all< is_ignored_and_final< Rules, Selector >::value... > {};

template< analysis::rule_type Type, unsigned Count, typename... Rules, template< typename... > class Selector >
struct is_final_node_2< analysis::counted< Type, Count, Rules... >, Selector >
   : is_all< is_ignored_and_final< Rules, Selector >::value... > {};
```

will select more rules which are "final".

The way this happens is that in contrast to `is_final_node`, which is not templated on `Rules...`, `is_final_node_2` will accept more rules such as `seq`, `sor`, `star`, and `plus` that have `analyze_t` which are templated on `Rules...`. However, `is_final_node_2` will only have `true` value if every Rule in `Rules...` is ignored (i.e. not selected) and final (i.e. a "simple" rule satisfying `is_final_node`). The fact that all of the `Rules...` must all be ignored is not actually a limitation here (relative to the previous optimization, which is to say that this optimization will capture a superset of the rules which would have resulted in Node allocation otherwise) because this optimization refers to only the more complex "parent" rules, and the "simple", "child" rules which are ignored will still not result in Node allocation. Is my understanding of this correct?

Given this interpretation, I might suggest that `is_final_node_2` be renamed to something like `is_final_parent_node`. Since this might technically be a "leaf" node though, maybe it makes more sense to rename `is_final_node`->`is_final_basic_node` and `is_final_node_2`->`is_final_node`. Given that they are not really examining Nodes either but rather Rules (which could produce a corresponding Node but due to this optimization will try not to do so), it might make sense to use the suffix `_rule` as well.

## Comments on the optimization

This is all a pretty clever and good compile-time optimization I think. My preliminary testing suggests that the runtime reductions for a simple calculator example are: 2-5% for a `true_type` selector, 25-45% for a more realistic selector, and 35-50% for a `false_type` selector.

However, these optimizations primarily apply to the case of a partial parse tree and to "simple" rules which are ignored / not selected. Rules which are not "simple" or rules which are still selected will still go down the path of creating Nodes on `start` which will not necessarily `match`. Thus, I think there are still more optimizations to be gained by using the `match` information only available at runtime.

I will investigate how to build such optimizations on top of your changes in `master`.

d-frey:
You understanding of my optimizations is correct and I've improved the names by using "leaf" in favor of other terms. Since all those helpers are internal, I also used one slightly longer and more descriptive name.

The `is_all`-implementation is just a minor detail and this little helper will go away in the future when we switch to C++17 or higher, being replaced with a fold expression anyways. For now, the extra 'bool' is no real issue and helps to prevent warning from some compilers for `Bs || true` (I've had that problem in the past when developing taocpp/sequences).

Your PR addresses further optimization opportunities and I'm looking forward to see an updated version.

d-frey:
It's been quite a while and the parse tree code has improved since, hence I wonder if we can close this issue or you have any remaining points that could be improved. As we are about to finish up the 3.0.0 release, now would be a good time to also get the parse tree finished for a first official (i.e. non-contrib) release. :)

PR #131, opened 2018-10-06
Add final argument for internal `parse_tree` state to `make_control` functions

mkrupcale:
Since we know that `make_control` is going to be given the internal state argument as its last argument from `parse_tree::parse`, we can explicitly take it as the function argument. This removes the need for using `tuple` methods and allows us to only pass on the user-provided state arguments to the underlying base `Control` class. Thus, any external `Action` or `Control` functions will also not have to accept the internal `parse_tree` state argument.

The `Control` `match` function is special, however, because it must use `make_control< Rule >::match` rather than `Control< Rule >::match` so that it can correctly invoke and pass the internal `parse_tree` state to the other `make_control< Rule >::(start|success|failure|raise|apply0|apply)` functions through `duseltronik`.

Note that this PR also results in performance improvements on the order of 5-10% for some simple `parse_tree` calculator benchmarks I've run. This is likely due to the reduced complexity by not using the tuple `tie` and `get` functions to create a tuple to extract the last variadic template argument.

### DIFF
--- a/include/tao/pegtl/contrib/parse_tree.hpp
+++ b/include/tao/pegtl/contrib/parse_tree.hpp
@@ -6,7 +6,6 @@
 
 #include <cassert>
 #include <memory>
-#include <tuple>
 #include <type_traits>
 #include <typeinfo>
 #include <utility>
@@ -253,6 +252,41 @@ namespace tao
             struct make_control< Node, Selector, Control >::control< Rule, false, true >
                : Control< Rule >
             {
+               template< typename Input, typename... States >
+               static void start( const Input& in, States&&... st, state< Node >& /*unused*/ ) noexcept( noexcept( Control< Rule >::start( in, st... ) ) )
+               {
+                  Control< Rule >::start( in, st... );
+               }
+
+               template< typename Input, typename... States >
+               static void success( const Input& in, States&&... st, state< Node >& /*unused*/ ) noexcept( noexcept( Control< Rule >::success( in, st... ) ) )
+               {
+                  Control< Rule >::success( in, st... );
+               }
+
+               template< typename Input, typename... States >
+               static void failure( const Input& in, States&&... st, state< Node >& /*unused*/ ) noexcept( noexcept( Control< Rule >::failure( in, st... ) ) )
+               {
+                  Control< Rule >::failure( in, st... );
+               }
+
+               template< typename Input, typename... States >
+               static void raise( const Input& in, States&&... st, state< Node >& /*unused*/ )
+               {
+                  Control< Rule >::raise( in, st... );
+               }
+
+               template< template< typename... > class Action, typename Input, typename... States >
+               static auto apply0( const Input& in, States&&... st, state< Node >& /*unused*/ ) noexcept( noexcept( Control< Rule >::template apply0< Action >( in, st... ) ) )
+               {
+                  return Control< Rule >::template apply0< Action >( in, st... );
+               }
+
+               template< template< typename... > class Action, typename Iterator, typename Input, typename... States >
+               static auto apply( const Iterator& begin, const Input& in, States&&... st, state< Node >& /*unused*/ ) noexcept( noexcept( Control< Rule >::template apply< Action >( begin, in, st... ) ) )
+               {
+                  return Control< Rule >::template apply< Action >( begin, in, st... );
+               }
             };
 
             template< typename Node, template< typename... > class Selector, template< typename... > class Control >
@@ -261,18 +295,16 @@ namespace tao
                : Control< Rule >
             {
                template< typename Input, typename... States >
-               static void start( const Input& in, States&&... st )
+               static void start( const Input& in, States&&... st, state< Node >& state )
                {
                   Control< Rule >::start( in, st... );
-                  auto& state = std::get< sizeof...( st ) - 1 >( std::tie( st... ) );
                   state.emplace_back();
                }
 
                template< typename Input, typename... States >
-               static void success( const Input& in, States&&... st )
+               static void success( const Input& in, States&&... st, state< Node >& state )
                {
                   Control< Rule >::success( in, st... );
-                  auto& state = std::get< sizeof...( st ) - 1 >( std::tie( st... ) );
                   auto n = std::move( state.back() );
                   state.pop_back();
                   for( auto& c : n->children ) {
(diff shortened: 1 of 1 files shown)


d-frey:
Nice! I wasn't aware that this is even allowed, and it seems to work on *most* platforms. Sadly, VS2015 gives an ICE. Any chance you can figure out how to make it work? Otherwise the alternative would be to drop support for VS2015. I'm not a Windows-guy, so I'm not sure if this would be acceptable.

mkrupcale:
So I'm not a Windows guy either, but based on the logs from the failed VS2015 Appveyor results, it looks like it's complaining about [`parse_tree.hpp:324`](https://github.com/mkrupcale/PEGTL/blob/3336ce245908a4a12d4b5d4cd06b470485bdde9e/include/tao/pegtl/contrib/parse_tree.hpp#L324) and [`parse_tree.hpp:379`](https://github.com/mkrupcale/PEGTL/blob/3336ce245908a4a12d4b5d4cd06b470485bdde9e/include/tao/pegtl/contrib/parse_tree.hpp#L379), which are both trailing return types for the `auto` function return type specifier. If VS2015 doesn't support these trailing return types, that's unfortunate because this really makes things so much easier to read than the previous `return_type_apply` templated structs that were used[1].

[1] https://github.com/taocpp/PEGTL/commit/4c10d84cfc24c22b18b0a17d938c32bb2b6ba15a

d-frey:
That seems likely and it's a "reasonably small" price to pay (YMMV). Could you please add an appropriate commit on top so we can check if this solves the problem? Thanks :)

mkrupcale:
Yeah, I'll play around with it a little bit. MS claims[1] that VS2015 should support trailing return types, so I wonder if it has to do with this combination of `decltype` with the trailing return type or something. Is there a way to force only VS2015 tests to run to save time?

[1] https://msdn.microsoft.com/en-us/library/hh567368.aspx

d-frey:
I've seen things you people wouldn't believe. Attack ships on fire off the shoulder of Orion. I watched C-beams glitter in the dark near the Tannhäuser Gate. Integer not having a unary operator +. Single types in need of pack expansion. All those moments will be lost in time, like tears in rain. Time to upgrade.

More serious: It usually works, but some combination of things leads to the ICE. Replacing the trailing return type with a "manually" calculated return type is by far not the weirdest work around we have :D

mkrupcale:
Yeah, I'm still trying to come up with a reproducer to play around with the issue. Since I don't have a local VS2015 (or any VS for that matter) to test, I'm currently working with Compiler Explorer...which is obviously not the most efficient way to debug this, so I'm open to other suggestions.

mkrupcale:
Okay, so I've managed to reproduce[1] both of these sets of errors in Compiler Explorer with VS2015. The reproducer is far from minimal (it is the brute-force preprocessed example test `contrib_parse_tree.cpp`) but should allow us to at least play with this and get it working in theory... The first set of errors seemed to be resulting from the trailing return type, while the second set of errors (when using the templated struct with the correct return type) appear to be related to the `start`, `success`, `failure` methods expecting 3 arguments but only getting 2. Not yet sure why this happens on VS2015 only.

[1] https://godbolt.org/z/ByXyBf

d-frey:
Could VS2015 be confused with the variadic argument list followed by another parameter? Seing the two parameters but assigning them to the States&&... and then expecting a third parameter?

mkrupcale:
That seems like a possibility. I haven't thoroughly investigated it, but I'm currently testing something which actually does at least compile in Compiler Explorer: using `decltype(auto)` trailing return type. This is technically a C++14 feature[1], but VS 2015 supports it[2], so it's a hack that might get it to work on VS 2015.

[1] https://en.cppreference.com/w/cpp/language/auto
[2] https://msdn.microsoft.com/en-us/library/hh567368.aspx

d-frey:
I've talked to @ColinH about dropping C++11 compatibility and requiring C++17. Note: This is only for taocpp/json, not for the PEGTL.

The pain with C++17 and older compilers (including VS2015), some workarounds that make things more complicated/slower to compile for everyone, maintaining our own backports for `std::byte`, `std::string_view` and `std::optional`, the long CI-cycle, ... - overall it feels like it is about time. It will depend on whether we can use C++17 in production environments, but with devtoolset-7 we should be able to get GCC 7.2+ on RHEL6/7.

mkrupcale:
> I've talked to @ColinH about dropping C++11 compatibility and requiring C++17. Note: This is only for taocpp/json, not for the PEGTL.

Okay, I don't think PEGTL has to drop C++11 compatibility in any case (although it would make things easier). The most recent test was just a workaround for VS 2015, and as far as I can tell[1], there's no way to force VS 2015 to use only C++11 and not C++14 (I don't think any `/std` was specified for the compiler in the CI or on Compiler Explorer). The alternative method of using the templated struct containing the proper return type is another valid approach for VS 2015 which only relies on C++11 features.

Both methods, however, still leave us with the issue of VS 2015 not seeming to support the `States...` argument in the middle of the function declaration. Perhaps this could be worked around on VS 2015 using some other method to extract the final argument from `States...` like you did with the `tuple` functions or perhaps something equivalent but somehow more efficient.

[1] https://msdn.microsoft.com/en-us/library/mt490614.aspx

d-frey:
I just mixed up taocpp/json and the PEGTL - sorry for the confusion. Obviously, dropping C++11 support for the PEGTL is currently not on the table.

For the issue at hand, we could `#ifdef` the code and let VS2015-users suffer by not removing the final argument for them on those methods. It would only affect the parse tree anyways, not the PEGTL in general. And with the extra argument, the parse tree might still be usable with VS2015. Let's call that plan C in case you can't find a reasonable work around which works with VS2015.

Plan B would be using tuple-based work arounds, probably also `#ifdef`-ed to isolate the cost to VS2105. Something like:

```c++
template< std::size_t... Is, typename... States >
void foo_impl(std::index_sequence<Is...>, States&&... st)
{
   auto ref = std::tie(std::forward<States>(st)...);
   bar( std::get<Is>(ref)... ); // ignores the last argument in st
}

template< typename... States >
void foo(States&&... st)
{
   // States must not be an empty pack!
   foo_impl( std::make_index_sequence<sizeof...(States)-1>(), std::forward<States>(st)... );
}
```

Not very nice, but it might work with VS2015.

PR #132, opened 2018-10-08
Add final argument for internal `parse_tree` state to `make_control` functions: VS 2015 support

mkrupcale:
This is same as #131 but intended for the 2.7.x branch for C++11 / VS 2015 support. The latest attempt to build on VS 2015 is following your Plan B approach from the discussion in #131.

### DIFF
--- a/include/tao/pegtl/contrib/parse_tree.hpp
+++ b/include/tao/pegtl/contrib/parse_tree.hpp
@@ -6,7 +6,6 @@
 
 #include <cassert>
 #include <memory>
-#include <tuple>
 #include <type_traits>
 #include <typeinfo>
 #include <utility>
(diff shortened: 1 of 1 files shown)


mkrupcale:
Do you have any idea why this is failing to build in VS 2015 only in release mode but not debug mode? The warnings are of the form
```
  c:\program files (x86)\microsoft visual studio 14.0\vc\include\type_traits(1276): warning C4702: unreachable code [C:\projects\pegtl\build\src\example\pegtl\pegtl-example-abnf2pegtl.vcxproj]
```
and the corresponding actual errors of the form
```
  c:\program files (x86)\microsoft visual studio 14.0\vc\include\type_traits(1276): error C2220: warning treated as error - no 'object' file generated [C:\projects\pegtl\build\src\example\pegtl\pegtl-example-abnf2pegtl.vcxproj]
```

I'm not really sure where the "unreachable code" is though based on these warnings.

d-frey:
Unreachable code warnings are often related to Release-mode, as the compiler is trying harder to prove that some code can be removed. But I have no idea which code it is complaining about or what triggered it. The error messages from VC++ are not really helpful. :(

mkrupcale:
Okay, so I think everything is working now. The issue was in the `raise_impl` functions where I was `get`ting the internal state argument to silence unused parameter C4100 warnings was not reachable after invoking the `Control` `raise` functions. So baring any particular issues with the implementation details, I suppose this PR could be merged.

d-frey:
Thanks!

PR #142, opened 2019-01-14
Fix string view

zhihaoy:
`std::string_view` has no range constructor: https://en.cppreference.com/w/cpp/string/basic_string_view/basic_string_view

PS: The UWP tests do not run on CI, please fix.


d-frey:
Thanks.

About UWP: It seems there is no way to enable it on AppVeyor, so I'll disable UWP tests. If I'm wrong about this and it can be done, I'll be happy to receive a PR.

d-frey:
FYI: I removed the UWP flags and the build should be fixed now.

PR #145, opened 2019-01-15
Fix Minor Spelling Mistakes & Update Travis Linux Distribution

sanssecours:
This pull request fixes some minor spelling mistakes I found while learning about PEGTL. The PR also updates the Linux Distribution used by the Travis build jobs.

### DIFF
--- a/src/example/pegtl/calculator.cpp
+++ b/src/example/pegtl/calculator.cpp
@@ -249,7 +249,7 @@ namespace calculator
       }
    };
 
-   // A number is a non-empty sequence of digits preceeded by an optional sign.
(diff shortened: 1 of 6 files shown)


uilianries:
I've used xenial for my builds on Travis and it's very stable now.

src/example/pegtl/calculator.cpp
```diff
@@ -249,7 +249,7 @@ namespace calculator
       }
    };
 
-   // A number is a non-empty sequence of digits preceeded by an optional sign.
```
ColinH: The correct word is "preceded".

PR #147, opened 2019-01-30
Removed pegtl::two and pegtl::three to add pegtl::any_amount

Sygmei:


nshcat:
Isn't the name `any_amount` kinda confusing? I mean, it doesn't really match any amount of the child rule, it matches it exactly N times. Maybe a different name would be more concise?

Sygmei:
Yup, I know the naming is not really good but I couldn't find a better one, it was originally named "x" (like "match this character x times").
Maybe one of those could fit : 
- match_n_chars (a bit long)
- match_n (short version)
- n_times

d-frey:
The whole thing got me thinking and I guess there are more things to clean up.

My current plan is:

* Add `ascii::rep_string< N, Cs... >`, based on `ascii::string< Cs..., Cs..., ... >`.
* Implement `ascii::two<C>` and `ascii::three<C>` based on the new `rep_string`.
* The "fun" rule `forty_two` allows `Cs...`, which is ambiguous. Should it mean 42 times the string `Cs...` or, as it is now, mean 42 characters, each one has to be one of those `Cs...`. (That ambiguity is the reason why `two` and `three` only allow a single character, right?)
* ...hence: `forty_two` should also only allow a single character, not multiple characters.
* ...also: `rep_string` does not suffer from this ambiguity.
* ...and if needed, we could also add `rep_one` which allows a set of characters.
* To decide: How much of this do we need? Is it actually used in practice often enough? Isn't `rep< 42, one< 'a', 'b' > >` and `rep< 42, string< 'h', 'e', 'l', 'l', 'o' > >` good enough?
* What about `utf*::string`? The current implementation is quite inefficient, as we do know the byte sequence we are looking for. It should translate the Unicode-string we are looking for to a byte-sequence at compile-time, then use `ascii::string` internally, maybe?

Any thoughts?

d-frey:
I ran a couple of benchmarks and hence replaced the implementations of `two` and `three` by simply using `string< C, C >` and `string< C, C, C >` which are equally efficient.

`forty_two` is only meant as a "fun/joke" rule, so no need to change anything there.

Leaving the question about the generic rules, potentially:

* `rep_one< N, Cs... >` as a short-cut for `rep< N, one< Cs... > >`.
* `rep_string< N, Cs... >` as a short-cut for `rep< N, string< Cs... > >`.

The questions are:

* How often will this be used/needed in practice? (I can't think of a real-world use-case)
* Is there a significantly more efficient implementation available? If not, there is not reason for a new rule - the short-cut alone is not giving enough benefits.

For that reason I added a quick hack of `rep_string` to the contrib section. It might be more efficient, but I don't think it is worth the effort of making it an official rule.

Feel free to continue the discussion, but for now I consider the matter resolved.

PR #149, opened 2019-02-13
Fix Conan build on Windows

uilianries:
Hi!

This PR will fix the Appveyor broken job. Is it PEGTL fault? No!

Conan changed some feature, where it does not detect the compiler when building header-only packages, but CMake is able to find a default compiler and mess all.

The fix is simple, tell Conan to use the default compiler, Conan will tell to CMake what it should use. However, the package id is the same, because we removed all settings from there.

Regards!


.conan/test_package/CMakeLists.txt
```diff
@@ -4,9 +4,10 @@ cmake_minimum_required(VERSION 3.2.0 FATAL_ERROR)
...
 include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
-conan_basic_setup()
+conan_basic_setup(TARGETS)
 
 find_package(PEGTL REQUIRED CONFIG)
 
 add_executable(${PROJECT_NAME} test_package.cpp)
-target_link_libraries(${PROJECT_NAME} taocpp::pegtl)
+target_link_libraries(${PROJECT_NAME} CONAN_PKG::pegtl)
+set_target_properties(${PROJECT_NAME} PROPERTIES CXX_STANDARD 17)
```
d-frey: Why would setting the C++ standard be necessary here? The PEGTL is a header-only library that requires C++17, so C++17 should be pulled in from the previous line already, or am I misunderstanding how CMake works?


```suggestion

```

d-frey:
Independent of the comment above I'll wait for the CI cycle and, if it has no problems, I'll accept this PR. Thanks for working on this 👍

.conan/test_package/CMakeLists.txt
uilianries: I think you are right, lemme review the cmake recipe. If add some public property and export your target, so any project will be able to receive all configuration needed to be build. But if your project doesn't export anything and need some extra feature, like c++17, so your consumers will need to update their cmake files too.

uilianries: Done!

PR #153, opened 2019-03-01
require_apply and require_apply0 with tags

pleroux0:
An alternative to `require_apply` and `require_apply0` that doesn't deepen the instantiation depth.

### DIFF
--- a/include/tao/pegtl/require_apply.hpp
+++ b/include/tao/pegtl/require_apply.hpp
@@ -17,22 +17,7 @@ namespace tao
    namespace TAO_PEGTL_NAMESPACE
    {
       struct require_apply
-      {
-         template< typename Rule,
-                   apply_mode A,
-                   rewind_mode M,
-                   template< typename... >
-                   class Action,
-                   template< typename... >
-                   class Control,
-                   typename Input,
-                   typename... States >
-         [[nodiscard]] static bool match( Input& in, States&&... st )
-         {
-            nothing< decltype( Control< Rule >::template apply< Action >( std::declval< const typename Input::iterator_t& >(), static_cast< const Input& >( in ), st... ) ) >();
-            return TAO_PEGTL_NAMESPACE::match< Rule, A, M, Action, Control >( in, st... );
-         }
-      };
+      {};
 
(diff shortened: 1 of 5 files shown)


include/tao/pegtl/require_apply.hpp
```diff
@@ -4,14 +4,6 @@
 #ifndef TAO_PEGTL_REQUIRE_APPLY_HPP
 #define TAO_PEGTL_REQUIRE_APPLY_HPP
 
-#include <utility>
-
-#include "apply_mode.hpp"
-#include "config.hpp"
```
d-frey: This include is needed for `TAO_PEGTL_NAMESPACE`, likewise for `require_apply0.hpp`

src/test/pegtl/verify_file.hpp
```diff
@@ -57,7 +57,7 @@ namespace tao
                parse< file_grammar >( in );
                TAO_PEGTL_TEST_ASSERT( !"no error on opening non-existing file" );
             }
-            catch( const std::system_error& e ) {
+            catch( const std::system_error& /*unused*/ ) {
```
d-frey: AFAICT you can just remove `/*unused*/` completely, as clang-tidy will not complain in this case.

d-frey:
I now see why you thought it is easy, while I was thought it's way more complicated. Your patch simplifies things, but the error message is not really helpful.

The current `require_apply` gives you an error message based on the attempt to actually call `apply`. This means you see the parameters that are passed and the candidates. Well, on GCC at least, Clang's error message is less clear.

Your light-weight solution is worth considering, as you may leave the base classes in if you like and I'm not worried about the CT-overhead with your PR. I might still add the existing `require_apply`/`require_apply0` helpers to contrib with a different name to help with better diagnostics.

If you fix the other minor issues I commented on, I will accept the PR. Thank you!

d-frey:
I committed another improvement, now it combines your light-weight solution which does not increase the template instantiation depth with the error messages from the previous `require_apply`/`require_apply0` and I managed to improve those messages (especially the "instantiated from..." call stack part) as well. I'm quite happy with how things turned out! 😃

PR #156, opened 2019-03-22
Make includable as a uniform CMake dependency

pleroux0:
This makes the way PEGTL is included in the project transparent to the
consuming project. If PEGTL is included with `add_subdirectory`
subsequent calls to `find_package` will find the PEGTL version included
with `add_subdirectory`. If PEGTL is included with `find_package`
subsequent calls to `add_subdirectory` will use the PEGTL version
included with find_package.

Since CMake targets are global there exists no way for multiple versions
of PEGTL to exist within one CMake build. Therefore, a FATAL_ERROR is
signalled if an attempt is made to include multiple different versions
of PEGTL.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,24 @@ cmake_minimum_required(VERSION 3.8.0 FATAL_ERROR)
 
 project(pegtl VERSION 3.0.0 LANGUAGES CXX)
 
+# Dont re-include pegtl
+if(${PROJECT_NAME}_FOUND)
+  if(NOT ${PROJECT_NAME}_VERSION STREQUAL ${PROJECT_VERSION})
+    message(FATAL_ERROR "Multiple mismatched PEGTL versions")
+  endif()
+
+  return()
+endif()
+
+# Keep track of pegtl version
+set(${PROJECT_NAME}_FOUND TRUE CACHE BOOL "" FORCE)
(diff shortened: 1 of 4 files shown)


d-frey:
Thanks for the PR, I'll accept it when Uilian had a chance to review it and approves as well.

Since I'm not a CMake user myself (I'm still happy with an old-school Makefile): Would it make sense to enhance our documentation at `doc/Installing-and-Using.md` with a couple sentences and small examples for both `add_subdirectory` and `find_package`? If so, could you kindly add it to your PR? (If all CMake users already know what to do please ignore this request). Thanks!

pleroux0:
The idea came from [C++Now 2017: Daniel Pfeifer “Effective CMake"](https://www.youtube.com/watch?v=bsXLMQ6WgIk) talk and the resulting [discussions on standardising the approach](https://gitlab.kitware.com/cmake/cmake/issues/17735). Creating a dummy config file is IMO the path of least resistance as it works with older CMake versions and nothing special has to be done in the consuming packages.

@uilianries It already works with `conan workspace` and pure CMake.  If #https://github.com/conan-io/conan/issues/4808 gets addressed then it can be made to work with `conan editable` as well. 

I should get some time to add to the documentation later today.

d-frey:
Thank you for the PR and enhancing the documentation, I think it will help a lot of people :)

PR #157, opened 2019-03-26
fix: Don't assume test executable position

pleroux0:
CMake automatically handles executable targets. The location of the
test targets might be changed by die project including PEGTL with
add_subdirectory.


d-frey:
Shouldn't we do the same for the examples then?

pleroux0:
Examples are only built and never run in the CMake script. The tests are run during `make test`.

d-frey:
D'oh, of course! :)

PR #166, opened 2019-04-13
Enable version checks in find_package

wravery:
This makes it possible to conditionally add PEGTL as a submodule or use the installed version if it's already a major version match (and greater than or equal to the minor.patch version), e.g.:

```cmake
find_package(pegtl 3.0.0 QUIET CONFIG)
if(NOT pegtl_FOUND)
  # If a compatible version of PEGTL is not already installed, build and install it from the submodule directory.
  set(PEGTL_BUILD_TESTS OFF CACHE BOOL "Disable PEGTL tests")
  set(PEGTL_BUILD_EXAMPLES OFF CACHE BOOL "Disable PEGTL examples")
  set(PEGTL_INSTALL_INCLUDE_DIR ${GRAPHQL_INSTALL_INCLUDE_DIR} CACHE STRING "Override PEGTL include install directory")
  set(PEGTL_INSTALL_CMAKE_DIR ${GRAPHQL_INSTALL_CMAKE_DIR}/pegtl CACHE STRING "Override PEGTL cmake install directory")
  add_subdirectory(PEGTL)
endif()
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *~
 build
 private
+/.vs
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,10 @@ endif()
 # Make package findable
 configure_file(cmake/dummy-config.cmake.in pegtl-config.cmake @ONLY)
 
+# Enable version checks in find_package
+include(CMakePackageConfigHelpers)
+write_basic_package_version_file(pegtl-config-version.cmake COMPATIBILITY SameMajorVersion)
+
 # install and export target
 install(TARGETS pegtl EXPORT pegtl-targets)
 
@@ -64,5 +68,6 @@ install(EXPORT pegtl-targets
   DESTINATION ${PEGTL_INSTALL_CMAKE_DIR}
 )
 
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/pegtl-config-version.cmake DESTINATION ${PEGTL_INSTALL_CMAKE_DIR})
 install(DIRECTORY include/ DESTINATION ${PEGTL_INSTALL_INCLUDE_DIR})
 install(FILES LICENSE DESTINATION ${PEGTL_INSTALL_DOC_DIR})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,9 @@ endif()
 # Make package findable
 configure_file(cmake/dummy-config.cmake.in pegtl-config.cmake @ONLY)
 
+# Ignore pointer width differences since this is a header-only library
+unset(CMAKE_SIZEOF_VOID_P)
+
 # Enable version checks in find_package
 include(CMakePackageConfigHelpers)
 write_basic_package_version_file(pegtl-config-version.cmake COMPATIBILITY SameMajorVersion)


d-frey:
Thanks, in general this looks good to me (I'll wait for Uilian's review, then accept it).

The only question I am wondering about is how we can (in the long term) solve the duplicate version sources: Both the `CMakeLists.txt` and `include/tao/pegtl/version.hpp` contain the version as we do not force users to use CMake. I guess there still isn't any good solution for this, but your patch will make the situation neither better nor worse. Anyways, if you have any idea how to improve that part, please let us know, otherwise just ignore me :)

d-frey:
Oh, and @uilianries: Should we enhance the documentation with Bill's technique?

wravery:
I'd try CMake's `configure_file()` with a version.hpp.in file to generate version.hpp, and then either install that FILE back to the source directory or use the source directory as the DESTINATION so you can still include PEGTL from a cloned repo or release archive without requiring an explicit install on the target machine.

Back when I was using [libgraphqlparser](https://github.com/graphql/libgraphqlparser) I noticed something similar to this approach, but it was used to generate a [pkgconfig](https://github.com/graphql/libgraphqlparser/blob/a292268db85a4d87632ed334df026853ffa433fe/CMakeLists.txt#L132) file. It also determined the version based on the latest git [tag](https://github.com/graphql/libgraphqlparser/blob/master/cmake/version.cmake), so the version wasn't even set in the CMakeLists.txt, but I think that's a little too fragile. It assumes that you're always building from a clone of the repo, it won't work with a release archive.

pleroux0:
The default `write_basic_package_version_file` requires the bitness (64bit or 32bit) to match. It will cause issues with the Conan package if moving between 32bit and 64bit. The correct way would be to configure a `pegtl-config-version.cmake.in` to work as expected.

d-frey:
@pleroux0 What would be your suggestion? Leave the situation as-is? Because from my perspective whether I manually keep the two files in sync or if I need CMake + a commit of the new generated file as an addition step for a release doesn't seem to make much of a difference. (Note that I'm not using CMake for the development)

wravery:
@pleroux0 I think you could compensate for that by clearing `CMAKE_SIZEOF_VOID_P` before calling `write_basic_package_version_file`:

```cmake
# if the installed or the using project don't have CMAKE_SIZEOF_VOID_P set, ignore it:
if("${CMAKE_SIZEOF_VOID_P}" STREQUAL "" OR "@CMAKE_SIZEOF_VOID_P@" STREQUAL "")
   return()
endif()
```

I'll go ahead and update this PR if that works, I expect the generated file should have the `@CMAKE_SIZEOF_VOID_P@` replaced with an empty string and shortcircuit the architecture check.

@d-frey You could hook up a CI pipeline to update the generated files automatically, but you'd want to make sure it had run before triggering the release. Ideally the release process would also be a CI/CD pipeline which gets triggered by changes to the generated version files in `master` or one of the downlevel branches. Then you'd just update `CMakeLists.txt` and wait for the pipeline to trigger a release.

pleroux0:
@wravery Yes. That would also work.

@wravery @d-frey Alternatively CMake can just parse the `version.hpp` to determine what version it is.

uilianries:
@d-frey said:

> Oh, and @uilianries: Should we enhance the documentation with Bill's technique?

Yes, we can update the documentation with an example including PEGTL with a specific version

Also, we can create a new PR (don't mix the things) with `configure_file` updating version.hpp

PR #167, opened 2019-04-15
Read project version from version.hpp

pleroux0:
Modifies CMakeLists.txt to determine the version from the `version.hpp` file to avoid duplication.

Some relevant discussion in #166

### DIFF
--- a/.conan/build.py
+++ b/.conan/build.py
@@ -52,7 +52,7 @@ class BuilderSettings(object):
 
     @property
     def reference(self):
-        """ Read project version from CMake file to create Conan referece
+        """ Read project version from version.hpp file to create Conan referece
         """
         pattern = re.compile(r"#define TAO_PEGTL_VERSION \"(\d+\.\d+\.\d+)\"")
         version = None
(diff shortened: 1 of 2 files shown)


uilianries:
Why not creating version.hpp by PEGTL version in CMakeLists.txt?

Reading a file and extracting the version by regex looks much complex than using the recommended cmake flow ([cmake tutorial](https://cmake.org/cmake-tutorial/)).

pleroux0:
That is also an option, but someone still has to check it into the repository. In a pure CMake project, it can reside in the build directory like generated files usually do (just as in the tutorial).

It makes sense to me having the `version.hpp` as the source of truth so that it can be used as pure header only library (just copy the sources).

d-frey:
I like this patch as it seems "simple" and straight-forward to me.

* `version.hpp` is the single source of truth.
* **If** someone uses CMake, `CMakeLists.txt` takes the version from the above file.
* If someone **does not** use CMake, it still works out-of-the-box.
* No extra step (manual/automated) to sync two files.

Am I missing any downside?

d-frey:
> Why not creating version.hpp by PEGTL version in CMakeLists.txt?
> 
> Reading a file and extracting the version by regex looks much complex than using the recommended cmake flow ([cmake tutorial](https://cmake.org/cmake-tutorial/)).

The only problem with the CMake flow is that it assumes CMake is used. The PEGTL simply doesn't enforce users (even Colin and me) to use CMake. Support for CMake is nice, being forced to use it is... not so nice. Maybe I'm old-school, but as long as there is no downside to Paul's approach, I tend to merge the PR soon. :)

uilianries:
oh, I see. I have to agree, your approach sounds better since this project doesn't require cmake.

Thank you @d-frey @pleroux0

PR #172, opened 2019-06-12
Fix implicit sign conversion warnings in clang

robertcampion:
Compiling in clang results in warnings like these:

```
include/tao/pegtl/internal/endian_gcc.hpp:141:17: warning: implicit conversion changes signedness:
'unsigned short' to 'std::int16_t' (aka 'short') [-Wsign-conversion]
         return __builtin_bswap16( n );
         ~~~~~~ ^~~~~~~~~~~~~~~~~~~~~~
include/tao/pegtl/internal/endian_gcc.hpp:141:36: warning: implicit conversion changes signedness:
'const std::int16_t' (aka 'const short') to 'unsigned short' [-Wsign-conversion]
         return __builtin_bswap16( n );
                ~~~~~~~~~~~~~~~~~  ^
```

This is because the `__builtin_bswapnn` functions expect unsigned integers.  In the `convert` functions for signed integers they are being casted to unsigned ints and back.  I made this explicit with `static_cast` to avoid the warnings.

For consistency I also swapped the order of the last pair of functions (uint64_t and int64_t), since all other pairs are in the order signed then unsigned.

### DIFF
--- a/include/tao/pegtl/internal/endian_gcc.hpp
+++ b/include/tao/pegtl/internal/endian_gcc.hpp
@@ -45,7 +45,7 @@ namespace TAO_PEGTL_NAMESPACE::internal
    {
       [[nodiscard]] static std::int16_t convert( const std::int16_t n ) noexcept
       {
-         return __builtin_bswap16( n );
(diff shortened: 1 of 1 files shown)


ColinH:
@robertcampion Thanks!

PR #178, opened 2019-07-04
Add android build on travis

Bjoe:
I added the "old" android builds.
You can now decided if you will merge this to have something or you wait if I'm done.

I have following ideas/task/improvements:
- Add android arm64 build and test. Starting August 1, 2019, [Google Playstore only accept Apps with 64-bit support included](https://developer.android.com/distribute/best-practices/develop/64-bit).
- Try to improve Android build: only download NDK and at least once with SDK to have an emulator.
- Use a [docker with an already installed android SDK/NDK](https://cloud.docker.com/repository/docker/bojoe/cpp-android-ndk-build-env-ubuntu) environment

If you will merge my stuff, please squash my commits! :-)


d-frey:
Adding the old build must fail, as they do not support C++17 - which is a requirement of the master branch of the PEGTL. I'll wait till you can add newer Android builds that support C++17.

Bjoe:
> Adding the old build must fail, as they do not support C++17 ...

@d-frey I mean with "old build" the build behavior from travis like before. My idea is to include a docker with an already installed android NDK.

At the moment the travis android build did following:
Every time, when an travis build is started, they install an android SDK with an android NDK. Unfortunately it is not possible at the moment to say, "use this NDK version" ... this means, travis is downloading on every build the newest NDK. At the moment this is NDK r20. On September 2019 NDK r21 will be shipped. This means if you include a new feature on September 2019 and get an build error on android, it can be because of the new NDK or of your feature ... I think it is not a good idea to automatically updated to the newest NDK ... first you didn't know this and you didn't see this on the travis build.

C++17 ...

In NDK r20 ...
<pre>
clang++ --version
Android (5220042 based on r346389c) clang version 8.0.7 (https://android.googlesource.com/toolchain/clang b55f2d4ebfd35bf643d27dbca1bb228957008617) (https://android.googlesource.com/toolchain/llvm 3c393fe7a7e13b0fba4ac75a01aa683d7a5b11cd) (based on LLVM 8.0.7svn)
Target: x86_64-unknown-linux-gnu
Thread model: posix
</pre>
So C++17 should be fully supported in the NDKr20...

Btw. I will close this PR. It was [working on the old "trusty" distribution build](https://travis-ci.org/Bjoe/PEGTL/builds/554199897) .... but you already switch to the newest "xenial" distribution ... there the installation of the android SDK is failing. I will figured out ... give me some days.

Bjoe:
So, I "refactor" the android build. I didn't use anymore the approach from travis. I downloaded the sdk-tools-linux-xxxxxxxx.zip, extract the zip, execute sdkmanager and install the missing dev tools like ndk-bundle, configure the emulator and thats it.
Locally it works fine, but sometimes the android emulator "hangs" in travis .. let me see if this happens here also ...
There are some thinks to discusse:
There are five android builds .... four with armeabi-v7a and one with arm64-v8a ... It is too much or? For example [API Level 19 is only distributed](https://developer.android.com/about/dashboards/index.html) 6.9 % ...

d-frey:
I don't think 5 builds is too much, but that will depend on the overall time it takes to run a build. It currently looks like ~20 minutes per build, so we should definitely not run every possible version of Android. But we should probably include some more recent versions. Anyways, let's first focus on getting the basic build method working and stable, it seems the execution currently hangs and that is, of course, a problem. To speed things up, you might want to reduce the builds to one or two temporarily. We can later use the pattern to add more builds as needed. Thanks for working on this!

Bjoe:
Again, the problem is that [the android emulator often stuck](https://travis-ci.org/taocpp/PEGTL/jobs/558140048#L487) but [sometimes it works](https://travis-ci.org/Bjoe/PEGTL/jobs/557933075#L474) ... I will skip the test the first ...

Issue #146, opened 2019-01-21
Question: Using PEGTL without RTTI


nshcat:
Hello,
as part of my thesis I was planning to use the PEGTL inside of the codebase of `clang`, which creates some issues since RTTI is disabled in that codebase and is non-trivial to turn on.

I don't use the `parse_tree`  functionality from the PEGTL, so the only thing that does not compile seems to be the demangling routines defined in `/internal/demangle.hpp`. How essential are those to the inner workings of the PEGTL?

Thanks a lot in advance

ColinH:
The demangle functionality is only used for debugging and error messages. To prevent the core library from both using `demangle()` and throwing exceptions, you can create your own Control class (class template `my_control< Rule >` that inherits from `pegtl::normal< Rule >`) that redefines the `raise()` function to do whatever you want it to do without calling `demangle()` (and possibly without throwing exceptions). Does this help?

nshcat:
Thank you a lot, this certainly helped!

PR #179, opened 2019-07-11
support demangling on known compilers without rtti

ohanar:
This more properly addresses the question posed in #146.

Unfortunately, even pulling in the demangle header prevents compilation with gcc when rtti is disabled (msvc is more forgiving about this). This makes PEGTL pretty much unusable when working in an environment without rtti. This patch just adds the demangling functionality needed without relying on rtti using compiler specific functionality for gcc, clang, and msvc.

Note, this patch is against the 2.x branch.


d-frey:
Thanks!

PR #181, opened 2019-07-27
add missing virtual dtors

kraj:
Fixes clang warnings
/usr/include/c++/v1/memory:3710:5: error: destructor called on non-final 'examples::null_json' that has virtual functions but non-virtual destructor [-Werror,-Wdelete-non-abstract-non-virtual-dtor]
__data_.second().~_Tp();
^

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/src/example/pegtl/json_classes.hpp
+++ b/src/example/pegtl/json_classes.hpp
@@ -62,6 +62,7 @@ namespace examples
          : json_base( json_type::array )
       {
       }
+      virtual ~array_json() = default;
 
       std::vector< std::shared_ptr< json_base > > data;
 
@@ -87,6 +88,7 @@ namespace examples
            data( in_data )
       {
       }
+      virtual ~boolean_json() = default;
 
       bool data;
 
@@ -103,6 +105,7 @@ namespace examples
          : json_base( json_type::null )
       {
       }
+      virtual ~null_json() = default;
 
       void stream( std::ostream& o ) const override
       {
@@ -118,6 +121,7 @@ namespace examples
            data( in_data )
       {
       }
+      virtual ~number_json() = default;
 
       long double data;
 
@@ -183,6 +187,7 @@ namespace examples
            data( in_data )
       {
       }
+      virtual ~string_json() = default;
 
       std::string data;
 
@@ -199,6 +204,7 @@ namespace examples
          : json_base( json_type::object )
       {
       }
+      virtual ~object_json() = default;
 
       std::map< std::string, std::shared_ptr< json_base > > data;
 


d-frey:
I don't understand what Clang is complaining about. The classes are derived from `json_base` which *does* have a virtual dtor. So the derived classes do have a virtual dtor as well. There is nothing wrong with the code. I understand that you are not responsible for this warning, but it seems strange. Do you have any explanation for why the "warn" about this?

kraj:
> I don't understand what Clang is complaining about. The classes are derived from `json_base` which _does_ have a virtual dtor. So the derived classes do have a virtual dtor as well. There is nothing wrong with the code. I understand that you are not responsible for this warning, but it seems strange. Do you have any explanation for why the "warn" about this?

@d-frey it is detecting UB see [1] and [2]

[1] https://wiki.sei.cmu.edu/confluence/display/cplusplus/OOP52-CPP.+Do+not+delete+a+polymorphic+object+without+a+virtual+destructor

[2] https://bugs.eclipse.org/bugs/show_bug.cgi?id=484780

d-frey:
I understand the general idea, but it simply doesn't apply here! The base class in our example already **does** have a virtual destructor, see https://github.com/taocpp/PEGTL/blob/b9070aadd1b7489484b17917f459103bc505c457/src/example/pegtl/json_classes.hpp#L33

I also tried to replicate the problem on Godbolt but I could not trigger the warning.

Looking at your patch, clang-tidy even complains that the derived classes' virtual dtor should use `override`, meaning it does see the base classes virtual dtor!

Finally, I don't see how adding virtual dtors to the **derived** classes could possibly fix the problem if the base class wouldn't have a virtual dtor. If you would delete an object through a pointer to the base class and the base class wouldn't have a virtual dtor, it doesn't even matter if the derived class has a virtual dtor - it would be ignored anyways.

The only thing that could make sense to me is the permissions. Does changing the base class' dtor from `protected` to `public` fix the problem for you (without adding dtors to the derived classes)?

ColinH:
Isn't something backwards? The base class d'tor is `protected`, so it doesn't need to be `virtual` because we, i.e. `shared_ptr< json_base >`, never `delete` a pointer to the base class. We should remove the `virtual`, which might have only been added to shut up a warning in GCC that I reported as bug, and that IIRC has been fixed a while ago...

d-frey:
I think clang-tidy will complain if the base class has virtual member functions but a non-virtual dtor.

kraj:
340110292b35d367205953a59e7eab28e1f4a0bb fixes this as well. I was on 2.1.4 release which did not have this fix. We can close this PR

ColinH:
Ok. Now let's hope that clang-tidy gets fixed soon.

d-frey:
I doubt that clang-tidy will be "fixed" even if someone were to report it, as clang-tidy works in more local contexts than, say, clang-analyze and they will most likely insist on warning about a missing virtual dtor when other virtual functions are present - independent from who derives which classes and how those are handled/deleted. And presonnaly I agree with the general rule, if you have an exception to the rule it usually involves some serious mental gymnastics from the developer and if you *really* want to avoid the virtual dtor, you can still add `// NOLINT` to suppress the clang-tidy warning in those (few) exceptional cases.

ColinH:
Detecting that the d'tor is protected should be local enough?

skyrich62:
Access, (ie: private, public, protected) doesn't apply in such cases.
If the base class dtor is virtual, then all derived classes dtors are virtual as well, (as has already been stated), regardless of access. 

One of the Meyers rules states that if a function is virtual in a base class it should be explicitly marked virtual in all derived classes, (not for the compiler, but for humans). 

Similarly, AUTOSAR rule A10-3-1 states: "Virtual function declaration shall contain exactly one of the three specifiers: (1) virtual, (2) override, (3) final."  Additionally, rule A10-3-2 states: "Each overriding virtual function shall be declared with the override or final specifier."

The reasoning here is to 1) Clarify intent, and 2) Avoid redundancy.  Both "override" and "final" are only valid on virtual functions, and only in derived classes.

See also C++ Core Guidelines [11]: C.128 "Virtual functions should specify exactly one of virtual, override, or final."

Thus:
---- cut ----
class Base
{
public:
    virtual ~Base();
};

class Derived: public Base
{
public:
    ~Derived() override;
};
==== end ====
is the preferred style.

Clang-tidy may be using the Meyers guidance, (which predates C++11).

ColinH:
My point with `protected` is that if a base class' d'tor is protected then you can't "polymorphically" `delete` an object through a pointer to said base class, and therefore you don't need to make it `virtual`. Which is why I usually stick to "base class d'tors must be either `virtual` *or* `protected`". It's ok to require `override` on virtual functions in derived classes, my issue is that the d'tor shouldn't be virtual in the first place when it's protected...

skyrich62:
Ok, right.  D'tors **are** special in this regard.  Your point is valid -- thanks for the clarification. :)

PR #187, opened 2019-10-28
Fix Cygwin build

samhocevar:
The functions `_byteswap_ushort()` etc. are not directly available on Cygwin (and MSYS2) because even though it defines `_WIN32` it does not expose MSVC intrinsics by default.

The choice here is to either fall back to GCC builtins, or include `w32api/intrin.h`. The former seemed safer and less intrusive to me.


d-frey:
Thanks!

PR #206, opened 2020-05-19
Update documentation following the change of byte_in_line to be 1-based.

samhocevar:


ColinH:
Thanks, we are slightly lagging behind with the documentation updates, but hope to get everything consistent RSN.

d-frey:
How does everyone feel about renaming "byte_in_line" to "column"? Yes, it is not technically correct when you consider multi-byte encoding, e.g. UTF-8. OTOH, it is already confusing when "byte" starts with 0 (because it's an offset), while "byte_in_line" starts with 1. Opinions?

samhocevar:
I’m in favour for the reasons you mention, but also because changing the name will make it more likely that users will double check their existing code when migrating to PEGTL 3. I know I didn’t and I had a `match` function that relied on `byte_in_line == 0`…

skyrich62:
I'm in favor.
Perhaps something else can be done for multi-byte encodings in the future.

--Rich

PR #212, opened 2020-07-19
doc: Update deprecated example

jubnzv:
Update example in Rules-and-Grammars.md to use actual methods of buffer_input class.
This doc is deprecated since https://github.com/taocpp/PEGTL/commit/cd35d8bae7d73637da1f9c077b55921c6cfdf897.


ColinH:
Thanks!

PR #215, opened 2020-08-23
Fixes #211

### LINKED ISSUE #211: Compile error in file_reader.hpp on Windows, std::filesystem::path uses wchar_t* which cannot be passed to std::fopen expecting char*
On Windows 10, compiling with gcc on mingw / msys2, std::filesystem::path::c_str() returns a const wchar_t* while std::fopen takes a const char_t*.  This causes a compile error.

```
C:/Users/dferr/CLionProjects/PEGTL/include/tao/pegtl/internal/file_reader.hpp:29:46: error: cannot convert 'const value_type*' {aka 'const wchar_t*'} to 'const char*'
       if( auto* file = std::fopen( path.c_str(), "rb" ) )

```
You won't be able to fix it with a simple cast.  It will require a UTF-16LE aware string conversion.

Amphaal:
Quick fix based on https://stackoverflow.com/questions/11352641/boostfilesystempath-and-fopen to allow compilation on MinGW systems.

### DIFF
--- a/include/tao/pegtl/internal/file_reader.hpp
+++ b/include/tao/pegtl/internal/file_reader.hpp
@@ -26,7 +26,7 @@ namespace TAO_PEGTL_NAMESPACE::internal
       throw std::filesystem::filesystem_error( "_wfopen_s() failed", path, ec );
 #else
 #if defined( __MINGW32__ )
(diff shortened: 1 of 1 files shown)


ColinH:
Thank you!

PR #223, opened 2020-10-12
Add table of equivalent forms to Rule Reference

eyepatchParrot:
When starting with the S expression example, I wanted a quick reference guide to quickly get up to speed in terms of already familiar languages like PEG and regex, and in terms of rules I was already familiar with: `if_must` from `must`.

Every row in every table which maps between pegtl rules depends only on rows above it, and every rule in the s expression example is covered. The long lines provide space alignment between columns in markdown, but with the links gets excessive.

This is just describing what I had in mind, let me know what you think or if it's not suitable for the project / goes somewhere else.

### DIFF
--- a/doc/Rule-Reference.md
+++ b/doc/Rule-Reference.md
@@ -32,6 +32,7 @@ When at least one parameter is given, i.e. `seq< A >` or `seq< A, B, C >`, `R` i
 
 ## Contents
 
+* [Equivalence Tables](#equivalence-tables)
 * [Meta Rules](#meta-rules)
 * [Combinators](#combinators)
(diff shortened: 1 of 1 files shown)


d-frey:
Thank you for the idea, I think it might be useful, although I'll discuss with Colin (when time permits) which table should go where. Personally, I think the reference section should not be containing PEG -> PEGTL or regex -> PEGTL tables, they might be better suited for other parts of the documentation.

ColinH:
For now we've kept the combinator table, albeit on the "getting started" page, but have removed the other tables since they would need to be greatly extended in order to be really useful, which then creates its own set of issues.

PR #224, opened 2020-10-16
Auto generate src/test/pegtl/file_äöü𝄞_data.txt

delpinux:
The test file "src/test/pegtl/file_äöü𝄞_data.txt" is now
auto-generated and no more stored in the git repository.

The reason for this change is that some filesystems (old ones?) do not
support this kind of files. Thus it makes it impossible on these
filesystems to extract the source tree (from a git repository for
instance). This is the case since commit 80a876d22a.


ColinH:
Are there any systems that have a C++17 compiler but don't support Unicode filenames?

delpinux:
I was also really surprised @ColinH 
This file system is a (stupid) proprietary NFS that is used on my office network. It is very bad and probably very expensive... I have no control upon it, but obviously since work stations are Linux boxes, we have recent gcc or clang compilers running on them.

d-frey:
TBH, I'm not too happy with this PR. It adds quite some complexity to the build systems (both make and CMake) only to support a case that shouldn't even exist. Why even work on an NFS-mounted volume that doesn't support Unicode? While I understand that it poses a problem to you, I have to wonder how many users are actually affected by this kind of problem.

I *might* consider a less intrusive PR, though. Would it be possible to disable this test (maybe by an explicit env variable)? Or is the checkout itself broken if the filename is present at all?

delpinux:
Hello @d-frey
I totally agree that this patch is not very nice.
I meant to send it to you since this kind of problem could occur to other users of PEGTL.
Feel free to reject/close this PR.
I use `git subrepo` (https://github.com/ingydotnet/git-subrepo) to incorporate PEGTL into my project and it is easy for me to removed this file from my project after each update.
Again, you can reject this PR it is not a big deal for me and my project.
Keep up the good work, and thanks again for PEGTL.

delpinux:
To answer your question: the checkout itself is broken. git cannot create the worktree on this filesystem for this peculiar file.
Actually I made some tests (few month ago), it is the 𝄞 character which is problematic.

Best regards

ColinH:
Since you say it's not a big deal for you we have decided to reject this PR for now with the idea that we can always resurrect it in case more people report problems with the (intentionally) strange filename and can't easily find a work around. And: Thank you for going the extra mile of preparing a PR for your issue!

PR #231, opened 2020-11-17
Updated contrib/json so that it is not greedy.

kelvinhammond:
contrib/json no longer consumes the whitespace after the json value.

### DIFF
--- a/include/tao/pegtl/contrib/json.hpp
+++ b/include/tao/pegtl/contrib/json.hpp
@@ -58,8 +58,8 @@ namespace TAO_PEGTL_NAMESPACE::json
    struct value;
 
    struct array_element;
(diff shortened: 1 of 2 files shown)


kelvinhammond:
I needed this so that I could parse an expression that takes json value or an identifier followed by one or more whitespaces and then a word but the json value was consuming all of the whitespaces.

kelvinhammond:
Tests pass, master fails because of build errors in the json examples. I'm not sure how to fix it. @d-frey can you take a look at the `example::control` in the `json_build.cpp`, etc examples?

d-frey:
I committed a different approach to master, could you verify that it fixes your problem?

kelvinhammond:
Your changes fixed my issue. Thank you.

Issue #216, opened 2020-08-24
Fix or nix Android CI builds


d-frey:
Currently, the Andriod builds are broken due to using std::filesystem::path. On some non-Android builds this was fixed by adding `-llibstdc++fs`, on some builds it just worked. Android, however, seems to be too old in some cases and I wouldn't know how to add the above option (in case it is required and helps). Also, the version we are using now seem to be a bit outdated.

We therefore need to either fix (and possibly modernize the build) or I will have to remove them (and most likely Android is then broken).

Any help is appreciated, @Bjoe ;)

wravery:
Relevant discussion in an NDK issue: https://github.com/android/ndk/issues/609

It looks like support was added fairly recently in NDK r22.

ColinH:
Do you have experience with the NDK? We would greatly appreciate a PR that would fix things again.

Bjoe:
@d-frey Sorry for the late response... some terrible thinks happens on Aug. ... anyways

I'm surprise why the the android toolchain is using libstdc++ ... it is not anymore updated by the NDK. PEGTL should use libc++ instead.
Give me some day, I will create a PR to fix it.

Bjoe:
@ColinH @d-frey As @wravery mentioned. `std::filesystem` is supported at NDK r22. At the moment NDK r21 is released. So r22 is a future release.

See also NDK issue: android/ndk/issues/1272

NDK r21 is not yet fully support C++17 features.

d-frey:
@Bjoe We appreciate your help with our little project, but of course people are more important than machines, so please take all the time you need to help your family - the PEGTL can wait. I am looking forward to see you again in person in the C++ user group once the pandemic is hopefully dealt with in spring/summer next year. For now, I'll disable the CI jobs for Android. Once r22 becomes available, we can try again to create a new CI job for it. Thanks again for your support!

Bjoe:
@d-frey Thank you for condolence ... it's three months ago so we are getting back to normal...

I found some more information about the [NDK r22 release](https://github.com/android/ndk/wiki#release-schedule). So there exists a NDK "22.0.6917172-beta1" version.
There are also a [changlog](https://github.com/android/ndk/wiki/Changelog-r22) for this version.

I used this version locally to build PEGTL for android and it compiles ! :-) 

I try now to add this beta version in travis .. see https://travis-ci.org/github/Bjoe/PEGTL/builds/744958133

[...]

PR #233, opened 2020-11-23
Fix Android CI builds

Bjoe:
This is a proposal to fix failing android CI builds. See issue #216 

It uses a docker where the android NDK r22 beta 1 is installed to build PEGTL.

See also travis build: https://travis-ci.org/github/Bjoe/PEGTL/jobs/745332637

It needs maybe some "cleanup" regarding to the output.

Maybe some output of the current NDK/SDK version would be useful.


Bjoe:
Let me introduce some output like: NDK version etc and maybe some other improvement. Give me some days.I will notice you if I'm done.

Bjoe:
It is more or less done. 
See travis build: https://travis-ci.org/github/taocpp/PEGTL/jobs/747196414

There is "error" 
```
ls: cannot access '/opt/androidSdk/platforms': Operation not permitted
-- ANDROID_PLATFORM not set. Defaulting to minimum supported version
16.
```
I didn't understand yet why this happens, because when I use the docker "localy" I didn't get this errors ....
But at least, PEGTL builds without any build errors so it does not harm the build. It would be nice to merge this and I will investigate deeper what are the problems on travis. But I think you know that is a time consumption task :-)

Btw. I use the docker approach because:
* It is not always downloading and installing the NDK/SDK stuff so it should be faster as before?! Not 100% sure because unfortunately the docker image is also always downloading. I expected that they only downloaded once ...
* It will not break your build because of some changes (for example, maybe there change the path for cmake etc) in the NDK/SDK  or because of some now license agreements etc etc. 
* But the biggest reason is because of maintenance. If there are problems, I or you can take the docker and build PEGTL for android without installing some SDK/NDK locally on your machine etc. 
The command to build PEGTL localy for android is:
```
cd [TO_PEGTL_SOURCES]
docker run -it --rm -v `pwd`:/source bojoe/cpp-android-ndk-build-env-ubuntu:r22beta1 "cmake -H/source -B/home/developer/build -GNinja -DCMAKE_TOOLCHAIN_FILE=/opt/androidSdk/ndk/android-ndk-r22-beta1/build/cmake/android.toolchain.cmake -DANDROID_ABI=armeabi-v7a -DANDROID_PLATFORM=android-26 -DCMAKE_BUILD_TYPE=Release -DPEGTL_BUILD_EXAMPLES=OFF && cmake --build /home/developer/build --target all"
```
See also in `.travis.yml`
You can also "manually" build and debug the build in the docker:
```
docker run -it --rm -v `pwd`:/source bojoe/cpp-android-ndk-build-env-ubuntu:r22beta1 "/bin/bash"
```
or
```
docker run -it --rm -v `pwd`:/source bojoe/cpp-android-ndk-build-env-ubuntu:r22beta1 "cmake -H/source -B/home/developer/build -GNinja -DCMAKE_TOOLCHAIN_FILE=/opt/androidSdk/ndk/android-ndk-r22-beta1/build/cmake/android.toolchain.cmake -DANDROID_ABI=armeabi-v7a -DANDROID_PLATFORM=android-26 -DCMAKE_BUILD_TYPE=Release -DPEGTL_BUILD_EXAMPLES=OFF && /bin/bash"
```
You will find the sources in the docker in `/source` .... but you can change this with the docker parameter `-v` ....

When there is a new NDK out, I will create a new docker on the docker hub and "tag" it with the new version. See https://hub.docker.com/r/bojoe/cpp-android-ndk-build-env-ubuntu
The docker sources are here:
https://github.com/Bjoe/DockerImages  
if you see I build some more docker environment also for other project because I always take an docker to develop and build .... :-). So I can use the latest ore newest compiler for everything without to mess up my host system ;-)

d-frey:
Thanks @Bjoe !

Some points for the future (some of which you already mentioned):
* Show the output from the compile just like with other jobs.
* Run the tests.
* Move those jobs to GitHub Actions.
* Now that we have fallbacks for std::experimental::filesystem: Maybe older Android versions are now able to use the PEGTL again? Could you check, please?

For moving stuff to GitHub Actions: I've seen that conanio hat some images for Android. I have no idea if that helps, but in our main-ci.yml there is a link to https://github.com/conan-io/conan-docker-tools where they list Android images. If those images are not suitable, you could probably still use your image with a job in our GitHub Actions main-ci.yml.

Bjoe:
@d-frey 

> * Show the output from the compile just like with other jobs.
> 
> * Run the tests.
> 
> * Move those jobs to GitHub Actions.
> 
> * Now that we have fallbacks for std::experimental::filesystem: Maybe older Android versions are now able to use the PEGTL again? Could you check, please?

Yep, it is possible and that's my idea.

I created some task on my story board ... I will start soon ....see https://github.com/Bjoe/PEGTL/projects/1

Bjoe:
@d-frey 

> For moving stuff to GitHub Actions: I've seen that conanio hat some images for Android. I have no idea if that helps, but in our main-ci.yml there is a link to https://github.com/conan-io/conan-docker-tools where they list Android images. If those images are not suitable, you could probably still use your image with a job in our GitHub Actions main-ci.yml.

Ok. I also involved in the [hunter project ](https://github.com/cpp-pm/hunter/pull/316) and there we use also "GitHub Actions" .... let me see what I can found out... but step by step.
First, I will also update [PEGTL in hunter](https://github.com/cpp-pm/hunter/tree/master/cmake/projects/pegtl) then I can see what are working on Github Actions build :-)

PR #236, opened 2020-12-03
Automatically fallback to older versions of std::filesystem

wravery:
This is addressing the discussion in #216. Along with the build and code changes I created a GitHub Actions CI/PR workflow to test it on Ubuntu 16.04 and 18.04, with gcc 7 and 8 on each of them. The 2 versions of Ubuntu work more or less the same way, but the gcc 7 jobs fall back to `std::experimental::filesystem`. The gcc 8 jobs detect that they need to link with `-lstdc++fs`, but they still use `std::filesystem`.

We might be able to add another job to do an Android build and test out the Boost.Filesystem fallback, which would also fix the original problem with #216.


Bjoe:
> We might be able to add another job to do an Android build and test out the Boost.Filesystem fallback, which would also fix the original problem with #216.

Yes it is possible. If your branch is merged I can add in the "docker" container (I switched to docker because it is easier to maintain. See PR #233) a boost filesystem lib for android.

d-frey:
Awesome, thank you very much @wravery and @Bjoe! I like the approach you took for more filesystem compatibility and I will try and convert more jobs to GitHub Actions, this is a great start! And I am looking forward to getting the PR for Android in as well, things are progressing quite nicely! 😀 

One question: Does using Ubuntu 16.04 provide any benefit or additional testing? Or could we just drop it to speed things up? I've also seen in other repositories that they were using ubuntu-latest - just to be clear: I want to have CI tests for the different compiler versions, I don't care too much about the distro used. I used older Ubuntu versions on TravisCI just because the old compiler versions weren't available on newer Ubuntu.

wravery:
@d-frey I don't think using Ubuntu 16.04 in addition to 18.04 actually provides any value right now. Similarly, installing Boost.Filesystem doesn't help right now since neither version of gcc needs to fallback that far (and I'm not sure if earlier versions even implement enough of the rest of C++17 to make it worthwhile). But I left both of those in to figure out a pattern for how we could matrix the OS/toolchain and how we could pull in (and cache) Boost.Filesystem for another toolchain like Android. Since the matrix jobs run in parallel, it doesn't increase the build time much, whichever one takes the longest determines the total time (looks like 6 or 7 minutes on average).

So we could certainly simplify this workflow if all it does is test the std::filesystem fallbacks. But it could also be a starting point for a general purpose Actions CI workflow, targeting more platforms/toolchains. In that case I think it would be worth it to have the coverage on a range of platform versions and toolchains, matrixed like this, and with a mechanism to add Boost.Filesystem on platforms that need it.

d-frey:
I agree that we should leave in the tests for boost::filesystem, if only because some users simply might *want* to use Boost.Filesystem instead. We can still remove older Ubuntu-builds later if we want to.

For the general matrix style: I'm already playing with it, have a look at the other GitHub Actions YML I committed. It's still work-in-progress, but it already works great on a variety of systems/compilers/...

There are a few things I don't yet understand:

* Clang 10 & 11 seem to fall back to std::experimental::filesystem. Is this because the image for them contains an older version of libstdc++?
* With Windows, it seems that I can only use the newest Visual Studio and only with ClangCL.

Any hints on these issues would be appreciated. 😄

wravery:
> Clang 10 & 11 seem to fall back to std::experimental::filesystem. Is this because the image for them contains an older version of libstdc++?

Maybe there's another library it should try linking with in the `OPTIONAL_LIBS` parameter to `check_filesystem_impl`? I'm using Clang 10 on my own Ubuntu 20.04 install, and it doesn't do this, it builds with `std::filesystem` and no other libraries. That's one of the reasons I started with the CI workflow, so I could trigger builds on the older platforms and confirm the fallbacks were working.

> With Windows, it seems that I can only use the newest Visual Studio and only with ClangCL.

It looks like VS 2017 is installed on `windows-2016`, it's not listed in the [supported runners](https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources) table, but there's a link to it in the [https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners#supported-software](supported software) section below that. I'm guessing at the runner's image name based on the version and the alias `windows-2019` that `windows-latest` has.

I don't see any runs where it didn't specify `-T ClangCL` was that in the matrix at some point?

d-frey:
On Ubuntu 20.04 the GCC version is newer, hence the newer libstdc++. Ubuntu 18.04's standard GCC is v7, so that makes sense. That problem will probably go away once 20.04 becomes the standard (which will probably break some older builds?)

For Windows builds you are right, I've made some changes. Let's see how that goes.

Do you know how to properly convince CMake to run sanitizers, clang-analyze, etc.? Or should I try to create jobs that use just make as I did before with TravisCI?

d-frey:
Nice, the Windows build is looking much better now. I just replaced AppVeyor with the new GitHub Actions completely and I cleaned up the TravisCI jobs. There are still many of those "special" jobs left, let's see if we can manage to replace all Travis jobs with GitHub Actions, but so far the experience with GitHub Actions is just awesome :)

Maybe you have more experience: What is best practice with GitHub Actions? One big YML or multiple smaller ones? Also, I feel like I created a bit of copy-paste mess especially in the steps:-section. Again, any help with this (or converting the remaining TravisCI jobs) is appreciated!

wravery:
> On Ubuntu 20.04 the GCC version is newer, hence the newer libstdc++. Ubuntu 18.04's standard GCC is v7, so that makes sense. That problem will probably go away once 20.04 becomes the standard (which will probably break some older builds?)

Ah, that makes sense. It's kind of cool that the build healed itself when there was an incompatibility I didn't even predict. :)

> Maybe you have more experience: What is best practice with GitHub Actions? One big YML or multiple smaller ones?

I actually started looking at GitHub Actions for the first time about 24 hours ago, so yeah, I agree it's really quick to pick up and the experience so far is great. :+1: I've mostly been using Azure DevOps, and that works pretty well when the code is in an ADO repo, but this is much smoother for GitHub projects. I'm going to switch all of my GitHub pipelines over to Actions as well.

IMO, best practice would be to have smaller workflows without duplication, but as much as possible they should be de-duplicated with `matrix` strategies. So if the `matrix` becomes unmaintainable, or if there's a lot of special purpose logic for a given target (e.g. using apt-get for Ubuntu), then probably better to split them. If the body of each `job` is the same except for some replaceable tokens (which seems to be the case right now) then it's better to de-duplicate them.

I think some of the `steps` could be de-duplicated some more using a matrix of OS, generator/compiler, and build type/config. The trick would be creative use of `exclude:` and `include:` properties to minimize the cross products where they don't apply. For instance, Windows doesn't build with gcc, and Linux doesn't build with MSVC. So maybe instead of putting either of those compilers in the shared matrix, include those compilers on those legs of the matrix: [including additional values](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-including-additional-values-into-combinations).

For per-platform setup steps, like setting up Xcode, you could use the [steps.if](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsif) property to only run it on a matching `matrix.os`. The one that I don't see any way to put in a `steps.if` condition is `container`. I don't know if you can use `matrix.include` to only add the `container` to `ubuntu-latest` runners, but that might be a way to handle it. You could also reduce the coverage to the set of compilers that come preinstalled on `ubuntu-latest` running directly on the job runner, and split the `container` version into a separate workflow for additional coverage.

d-frey:
I've added GCC 7 and GCC 8 to the Main CI jobs, the fallbacks are working fine. Since std-filesystem-fallbacks.yml is currently testing the same: Would it make sense if you convert it into a test for the Boost.Filesystem fallbacks? Those are currently untestet and you already install Boost.Filesystem there.

d-frey:
There is one more issue that I don't understand: Some jobs sometimes fail, e.g. https://github.com/taocpp/PEGTL/runs/1499774885?check_suite_focus=true - the compiler itself aborts with strange problems like "double free or corruption" or "aborted by signal". Do you (or anyone else) have any idea what is going on there?

wravery:
That's very odd, it looks like a compiler bug to me, but maybe since it's intermittent it's related to memory pressure? When I tried adding the `-j` flag to `CMake` to parallelize the builds I noticed I got a few failures which indicated that it had exhausted virtual memory, so I rolled that back. Maybe it needs a [strategy.max-parallel](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymax-parallel) cap to run fewer jobs at once, although that'll slow down the overall turnaround time. It sounds like GitHub already optimizes for the number of available VMs.

wravery:
There are also [timeout-minute](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes) options on both the individual `steps` and the `job` itself. It mentions that the `job` defaults to a 6 minute timeout, but the total runtime for that failure you linked was less than 4 minutes, so that seems unlikely.

PR #237, opened 2020-12-06
Add a CMake option to force building with Boost.Filesystem

wravery:
I added an option to CMakeLists.txt called `PEGTL_USE_BOOST_FILESYSTEM`, which defaults to `OFF`. If enabled, it will bypass all of the `std::filesystem` auto-detection and behave as though it fell all the way back to Boost.Filesystem.

Doing that exposed some incompatibilities between the code which was still using `std::error_code` and `boost::filesystem::filesystem_error`, which expected a `boost::system::error_code`. So I needed to make `internal` namespace versions of a few more of those types and replace at least one `catch (const std::system_error& )` statement with a more specific `catch (const internal::filesystem::filesystem_error& )` in the tests. Normally `std::filesystem::filesystem_error` inherits from `std::system_error`.

I also switched `std-filesystem-fallbacks.yml` over to just testing with `PEGTL_USE_BOOST_FILESYSTEM`. While testing that, I noticed that the armv7 tests consistently failed in the `Main CI` workflow even though it built successfully. Since we can't run armv7 test binaries on an x86 runner, I added an `if` condition to the test step which suppresses the test runs if the image contains `armv7`. The `mingw` tests also seem to fail consistently, but I don't know if that's surmountable or not, so I left them as is.

### DIFF
--- a/.github/workflows/main-ci.yml
+++ b/.github/workflows/main-ci.yml
@@ -48,7 +48,7 @@ jobs:
       run: cmake --build . --config ${{matrix.build_type}}
 
     - name: Test
-      if: ${{!contains(matrix.image, "arm7")}}
+      if: ${{!contains(matrix.image, 'arm7')}}
       working-directory: build/
       run: ctest -C ${{matrix.build_type}} --output-on-failure
 
--- a/.github/workflows/main-ci.yml
+++ b/.github/workflows/main-ci.yml
@@ -48,7 +48,7 @@ jobs:
       run: cmake --build . --config ${{matrix.build_type}}
 
     - name: Test
-      if: ${{!contains(matrix.image, 'arm7')}}
+      if: ${{!contains(matrix.image, 'armv7')}}
       working-directory: build/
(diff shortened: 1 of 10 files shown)


d-frey:
Thank you very much!

I missed why the ARM tests failed, but it makes sense. However, I will remove those builds now as I don't see much value in them when we can't run the tests.

For MinGW and the GCC x86 test I hope @uilianries can have a look as those images are provided by Conan and there errors are just weird.

Now we just need to add some documentation for the filesystem fallbacks, including the option to choose Boost.Filesystem. Would you like to write something up or shall Colin and me take care of it?

d-frey:
I refactored the CI files, the Boost.Filesystem jobs are now integrated with the other Linux jobs in `linux.yml`. I *might* have gone a bit overboard with adding more jobs for all GCC/Clang versions and Debug/Release. I'll probably dial it down a bit in the future, I just wanted to make sure that we don't miss anything here.

PR #240, opened 2020-12-14
Android build: Show the output from the compile just like with other jobs

Bjoe:
First task from https://github.com/taocpp/PEGTL/pull/233#issuecomment-739147910 
* Show the output from the compile just like with other jobs


d-frey:
Thanks!

PR #241, opened 2020-12-14
Android build on GitHub Actions

Bjoe:
Move android build from travis CI to GitHub Actions


.github/workflows/android.yml
```diff
@@ -0,0 +1,32 @@
...
+        build_type: [Debug, Release]
+        android_abi:
+          - armeabi-v7a
+          - arm64-v8a
+        android_platform:
+          - android-26
+
+    runs-on: ubuntu-latest
+
+    container:
```
Bjoe, line 21: @d-frey Where can I find the documentation on using the docker on GitHub-Actions ? ... I didn't find one :-( I need to set some "parameters" on docker "run"...

d-frey: I'm not experienced with any of that, but maybe this might help: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer

d-frey: Also, check out our linux.yml, it is using containers from Conan where we also need to set some options.

Bjoe: > but maybe this might help: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer

Yep, this was the right hint :-) many thanks. 

> Also, check out our linux.yml, it is using containers from Conan where we also need to set some options.

Yep I know. A copy from the linux.yml and past it in android.yml :-)

Bjoe:
Please squash my commits.

d-frey:
Thank you very much, this is already very helpful. There are always things to improve and I'm looking forward to those, but having migrated everything to GitHub Actions is a big step. I'll follow up with adapting the README to the new situation :)

d-frey:
FWIW, I also enabled building the examples which seems to work just fine.

PR #242, opened 2020-12-17
Wrap filesystem paths in quotes

wravery:
The [cppgraphqlgen](https://github.com/microsoft/cppgraphqlgen) project includes PEGTL as a submodule and looks for it with a minimum version number in `find_package`, so if you have a recent enough version installed on the system (e.g. with vcpkg) it will prefer the system version, but it will fall back to using `add_subdirectory` for the submodule if it's missing or not recent enough. 

I noticed when I bumped the submodule and CMake version number for PEGTL in cppgraphqlgen that I started getting CMake errors in `pegtl-config.cmake`.  VS 2019's default build config on my system is called "x64-Debug (default)", which has a space in it. When PEGTL configures the `pegtl-config.cmake` file, it doesn't escape or quote the filesystem paths, so the build directory that VS 2019 creates with that config name using a space by default breaks.

### DIFF
--- a/cmake/dummy-config.cmake.in
+++ b/cmake/dummy-config.cmake.in
@@ -2,4 +2,4 @@
 # When a dependency is added with add_subdirectory, but searched with find_package
 
 # Redirect to the directory added with add_subdirectory
-add_subdirectory(@PROJECT_SOURCE_DIR@ @PROJECT_BINARY_DIR@)
+add_subdirectory("@PROJECT_SOURCE_DIR@" "@PROJECT_BINARY_DIR@")


d-frey:
Thanks. Are there other places where we'd need to add quotes? For example when a user checks out the PEGTL and tries to build tests and examples with VSCode or VS? (I don't use either of those)

wravery:
I haven't found any, but CMake `configure_file` and any shell scripts seem like candidates. Thinking about it some more, it should probably also run through `file(TO_CMAKE_PATH)` to make a portable path format (using / separators in particular). If it has any escaped characters it should probably double-escape them before calling `configure_file` since that'll just replace the token without any other escaping (unless there's a way to do that with an option on `configure_file` I don't know about).

PR #245, opened 2020-12-19
parse_tree::basic_node now uses template Source, defaults to string_view

### LINKED ISSUE #243: Improve parse_tree::basic_node memory usage
Is there any reason why `basic_node` copies `source`?

barbieri:
There is no reason to copy the string over and over again, most of the
time the source will be alive and a string_view will be enough.

If there is any case where this is not the case, then `basic_node` can
be told to use a different type via the second template parameter.

Closes #243

### DIFF
--- a/include/tao/pegtl/contrib/parse_tree.hpp
+++ b/include/tao/pegtl/contrib/parse_tree.hpp
@@ -32,7 +32,7 @@
 
 namespace TAO_PEGTL_NAMESPACE::parse_tree
 {
-   template< typename T >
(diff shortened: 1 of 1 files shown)


d-frey:
There are sometimes reasons why you might want to copy the source, e.g. when the input gets destroyed but you want to keep the nodes around. But since you PR allows to change that if needed, I'll accept the PR. Long term, the whole input handling and especially the source needs an overhaul, but that will result in a new major version.

Issue #244, opened 2020-12-19
Improve parse_tree::basic_node is_type


barbieri:
Is there any reason why the demangled version is compared directly?

Comparing `typeid().hash_code()` could improve that, you can compare the demanged version only if the hash code is different.

PR #246, opened 2020-12-19
parse_tree::basic_node check hash_code() before demangled string

barbieri:
typeid().hash_code() guarantees the same type will have the same
hash_code. Since it's fixed size (size_t) and much faster/simpler to
compare, use it in `is_type()`.

However, different types can have the same hash_code(). It is not
common, but is possible, then we must compare the strings afterwards.

Closes: #244

### DIFF
--- a/include/tao/pegtl/contrib/parse_tree.hpp
+++ b/include/tao/pegtl/contrib/parse_tree.hpp
@@ -39,6 +39,7 @@ namespace TAO_PEGTL_NAMESPACE::parse_tree
       using children_t = std::vector< std::unique_ptr< node_t > >;
       children_t children;
 
+      size_t type_hash;
       std::string_view type;
       std::string source;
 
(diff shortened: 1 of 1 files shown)


barbieri:
hum...  linux-gcc-extra (-fno-rtti, Debug) failed, since there is an explicit `no-rtti` I believe this is a no go. Is there any fast yet non-rtti to fix this?

Since `demangle()` is constexpr, maybe we could compute a hashcode based on it, making it constexpr as well?

```
In file included from /__w/PEGTL/PEGTL/src/test/pegtl/contrib_parse_tree_to_dot.cpp:6:
/__w/PEGTL/PEGTL/include/tao/pegtl/contrib/parse_tree.hpp: In member function 'bool tao::pegtl::parse_tree::basic_node<T>::is_type() const':
/__w/PEGTL/PEGTL/include/tao/pegtl/contrib/parse_tree.hpp:71:40: error: cannot use 'typeid' with '-fno-rtti'
   71 |          return type_hash == typeid( U ).hash_code() && type == demangle< U >();
      |                                        ^
```

d-frey:
As you have dicovered, `-fno-rtti` is a feature. But I committed a fix as the pointers are, realistically, the same if it is the same type. And if they are not, we fall back to comparing the string itself.

PR #254, opened 2021-01-18
peg2pegtl example

redmercury:
This change implements a pegtl generator from peg grammars, as defined [in the original peg paper](https://pdos.csail.mit.edu/~baford/packrat/popl04/peg-popl04.pdf).  Included is the original peg grammar (peg.peg).  I've regenerated the peg pegtl grammar components in peg.hpp with the output of peg.peg, and it is able to self-host.

### DIFF
--- a/include/tao/pegtl/contrib/peg.hpp
+++ b/include/tao/pegtl/contrib/peg.hpp
@@ -1,0 +1,192 @@
+// Copyright (c) 2021 Daniel Deptford
+// Please see LICENSE for license or visit https://github.com/taocpp/PEGTL/
+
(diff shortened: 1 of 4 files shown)


d-frey:
* I fixed the namespace issue (`tao::pegtl::` could be redefined to something else via `TAO_PEGTL_NAMESPACE` and you don't need it anyways)
* I used some more idiomatic rules like `list<>` and `until<>` and I removed some superfluous `seq<>`s. It's not exactly the direct mapping from the paper now, but we shouldn't show such inefficient grammars in the examples. I hope you agree :)

PR #264, opened 2021-04-23
Documentation: change src/examples/ to src/example/

jbomanson:
Hello

I just started trying out pegtl and it is working great so far!

In the process of trying it, I spotted a few file paths in the documentation that referred to `src/examples`, which does not exist. I fixed the paths to `src/example/` in this pull request.


ColinH:
Thank you!

PR #268, opened 2021-08-19
fix dec_octet in contrib/uri grammar

avsej:
The grammar for dec_octet does not conform RFC, it should not allow any
octet, but rather [0-255].

### DIFF
--- a/include/tao/pegtl/contrib/uri.hpp
+++ b/include/tao/pegtl/contrib/uri.hpp
@@ -32,7 +32,11 @@ namespace TAO_PEGTL_NAMESPACE::uri
    using colon = one< ':' >;
 
    // clang-format off
-   struct dec_octet : maximum_rule< std::uint8_t > {};
+   struct dec_octet : sor<
+      rep_min_max< 1, 2, abnf::DIGIT >,
+      seq< one< '1' >, rep< 2, abnf::DIGIT > >,
+      seq< one< '2' >, range< '0', '4' >, abnf::DIGIT >,
+      seq< one< '2' >, one< '5' >, range< '0', '5' > > > {};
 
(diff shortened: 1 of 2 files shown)


d-frey:
That is not the right solution. The `maximum_rule< std::uint8_t >` already only accepts 0-255. The problem is that it throws an exception on overflow instead of failing locally (by returning `false`). I'll discuss the semantics of `contrib/integer.hpp` with @ColinH when we have time, but I guess that is exactly the reason why it is still in `contrib`.

ColinH:
I believe we had already started to look into the `throw` vs. `false` question without any clear result. Looking at the source in question one could make a case to change `match_and_convert_unsigned_with_maximum()` to not throw, and then possibly change everything else, too, in line with delivering just the basic building blocks and the expectation that the user put together their own actions with exceptions if and when necessary.

Regarding the original issue, I just tested inputs with invalid IPv4 addresses, and things like `256.1.1.3` or `1.1.1.300` are correctly rejected (if, as outlined above, with an exception instead of a return value). @avsej Which input led you to believe that invalid inputs are accepted?

d-frey:
@ColinH The problem is an input like `http://256bytes.com` - the exceptions prevents other options to be considered.

ColinH:
Right, because hostnames can begin with digits. And, worse, they can even be only digits and dots, so `300.300.300.300` is a valid hostname via the `reg_name` production of RFC 3986. It's easy to forget since nobody sane does that...

I've now changed `maximum_rule<>` to return `false` on overflow which is more in line with what one would expect from a pure matching rule. (The bigger question always was how to handle the combined action and rule case.)

The original issue should also be resolved by 707fd88b3add111258d088dcfdbf27a80dfbad3c.

PR #269, opened 2021-09-25
Fix proto3 example

hutorny:
This pull request suggests fixes in proto3 example.
With these fixes it successfuly parses 39 proto3 files from google and MAVSDK depositories


ColinH:
Thank you, looks good to me. You're probably the first person to really use this grammar. One thing you can simplify, the PEGTL already includes the `keyword<>` rule in `tao/pegtl/asci.hpp`. If you can remove your redundant definition I'll merge this PR.

hutorny:
Removed

ColinH:
Thanks again!

PR #272, opened 2021-10-01
Added IRI to contrib

kelvinhammond:


kelvinhammond:
Not sure what to set the copyright to, please advise.

ColinH:
Could the first `sor` be changed to use `ranges` instead? Otherwise looks good, except perhaps for some minor formatting that we can fix later.

kelvinhammond:
It could be but this formatted better, i have a version of this with that,
what do you prefer?

d-frey:
Using `ranges<...>` is more efficient than using `sor<range<...>,range<...>,...>`, so the former is preferred.

d-frey:
For the copyright: For now, just go with the current header - I'll change it later when hopefully all contributors agreed to the license change.

For the formatting: We'll take care of that, don't worry about it too much.

kelvinhammond:
Updated

PR #313, opened 2022-04-21
Make parse_tree Node::unwind optional

wravery:
Split into a separate branch/PR from #309.


d-frey:
I'm wondering if this code still has any benefits, as the `state.pop_back()` can not be eliminated and the only thing it now does is skip the call to the node's `unwind` method. But that method can be empty (which it is in the provided default node class), so the compiler will easily optimize it away anyways. Am I missing any benefits? Or is this the point, that you are able to skip the definition of an empty `unwind`-method for your custom node class at the expense of some compile-time overhead?

wravery:
> ...Or is this the point, that you are able to skip the definition of an empty `unwind`-method for your custom node class at the expense of some compile-time overhead?

Yes, that's the goal. Since unwind is typically optional for a Control type, it seems strange that the `parse_tree::Node` would require it to be implemented (even if it's empty by default).

PR #315, opened 2022-05-13
Workaround broken `__FUNCSIG__`

wravery:
I verified that my project works with this change against main. I'd love to get this backported to a 3.2.6 release as well, since that would be a smaller change for me to adopt in my project.


d-frey:
I fixed it without removing support for no-rtti. Can you confirm it works for you? If it does, I'll backport it to 3.x.

wravery:
> I fixed it without removing support for no-rtti. Can you confirm it works for you? If it does, I'll backport it to 3.x.

Yes, that works, thanks!

PR #318, opened 2022-07-06
Added support for fixed-length list.

qawbecrdtey:
I made a list which supports both dynamic and fixed length. It should be backward compatible.

Also, the ifdef was added in `rules.hpp` for demonstration purpose.
This should be removed after some discussion.

### DIFF
--- a/include/tao/pegtl/contrib/list_length.hpp
+++ b/include/tao/pegtl/contrib/list_length.hpp
@@ -1,0 +1,51 @@
+#ifndef PEGTL_LIST_LENGTH_HPP
+#define PEGTL_LIST_LENGTH_HPP
+
+#include "../internal/result_on_found.hpp"
+#include "../internal/rules.hpp"
+
+namespace tao::pegtl::list_length
+{
+
+   template< uint64_t N, typename Rule, typename Sep >
+   struct list_length;
+
+   template< typename Rule, typename Sep >
+   struct list_length< 0, Rule, Sep >
+      : success
+   {};
+
+   template< uint64_t N, typename Rule, typename Sep >
+   struct list_length
+   {
+      using rule_t = list_length;
+      using subs_t = type_list< Rule, Sep >;
+
+      template< tao::pegtl::apply_mode A,
+                tao::pegtl::rewind_mode M,
+                template< typename... >
+                class Action,
+                template< typename... >
+                class Control,
+                typename ParseInput,
+                typename... States >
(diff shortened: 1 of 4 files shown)


ColinH:
Thank you for making a PR, much appreciated! A list with a fixed length is a nice addition, however I think a simpler approach should be used, and I'm not sure about the name...

d-frey:
What exactly do you mean by "backward compatibility"?

qawbecrdtey:
Never mind, I just thought that both list of dynamic and fixed length should be implemented with the same struct.

d-frey:
We prefer divide-and-conquer. What we really need is a better name for this rule. Also, @ColinH, would it make sense (or is it a problem) to have a minimum and a maximum size of the list, where the default maximum is the minimum?

ColinH:
@qawbecrdtey Putting everything into the same class or template can be useful, but I can't see any benefit in this particular case. In particular with the fixed-length and open-ended list rules having different user-facing names, making the latter a dedicated special case of the former in the internal implementation does not seem natural.

ColinH:
@d-frey That leaves naming, I like `list_min_max` for the `rep_min_max`-based list but I'm having trouble coming up with a nice/better name for `list_with_size`.

d-frey:
I think we could drop 'list_with_size' and only keep `list_min_max` (with default max=min), as those rules are IMHO not required very often and should therefore only be in contrib anyways. If a user needs a more special version of it, they can simply create their own, as it's only aliases / small forwarders to existing rules.

ColinH:
That's one way to solve the naming question :-)  I like the idea in general, though there's one detail that might be an issue. A `list_min_max< N, N, ... >` based on `rep_min_max< N, N, ... >` does not behave like `list_with_size< N, ... >` based on `rep< N, ... >` because `rep_min_max` has a `not_at` to verify that there are no further repetitions while `rep` does not.

ColinH:
This whole subject is actually more complicated than it appeared at first glance. Quelle surprise. For example assuming `under = one< '_' >`, clearly `list_min_max< 1, 2, alpha, under >` should reject `"a_b_c"`, but what about `"a_b_"`? Depending on the surrounding grammar either matching or not matching will be a more appropriate choice. I'm now leaning towards removing both `list_with_size` and `list_min_max` again because in this light it feels like overall there are too many possibilities for "list with size" and it there is no canonical one-size-fits-all choice of semantics.

PR #325, opened 2022-11-29
Remove problematic language from the code

Cropi:
Hello,

I would like to replace problematic terms with alternatives that are appropriate to the particular project and context. 
Similar to https://github.com/taocpp/PEGTL/commit/731bc8ff92feb41a09efa31d7b8f74a8af132d77, rename:
- white-list to allow-list
- black-list to block-list


ColinH:
Thanks, if you could please change "a allow-list" to "an allow-list" then this can be merged.

Cropi:
Thanks for the quick review. I applied the change.

PR #343, opened 2023-04-04
Feature/propagate node

skyrich62:
While creating a parse tree, it would be nice to propagate the current node to the user's control methods, (e.g. start, success, failure).

This way the user has additional control over handling the nodes of the tree.

### DIFF
--- a/include/tao/pegtl/contrib/parse_tree.hpp
+++ b/include/tao/pegtl/contrib/parse_tree.hpp
@@ -292,7 +292,7 @@ namespace tao::pegtl::parse_tree
          {
             state.emplace_back();
             state.back()->template start< Rule >( in, st... );
-            auto &n = state.back()->children.back();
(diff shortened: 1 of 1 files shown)


include/tao/pegtl/contrib/parse_tree.hpp
```diff
@@ -290,9 +290,10 @@ namespace tao::pegtl::parse_tree
          template< typename ParseInput, typename... States >
          static void start( const ParseInput& in, state< Node >& state, States&&... st )
          {
-            Control< Rule >::start( in, st... );
             state.emplace_back();
             state.back()->template start< Rule >( in, st... );
+            auto& n = state.back()->children.back();
```
ColinH, line 295: AFAICT `children` can be empty here.

ColinH (changes requested):

PR #344, opened 2023-04-30
separated_seq for 2.x

almazgaliev:
i needed separated_seq for project which was written in c++11 standart, so i copied separated_seq from 3.x version and applied minimum fixes for it to work on c++11


ColinH:
The unit test compiles and executes successfully? Speaking of, the unit test `.cpp` file needs to be added to `CMakeLists.txt` in `src/test/pegtl`.

almazgaliev:
Sorry, forgot about adding it to CmakeLists.txt, just did that.

ColinH:
If it works in 3.x and, after a couple of changes, it works for you in 2.x that should hopefully be good enough for everybody. Thanks!

Issue #347, opened 2023-06-13
Any consideration of ghc::filesystem


baconpaul:
Hi!

I'm using tao json which uses pegtl filesystem support. Right now you check for std::fs std::experimental::fs and Boost::fs. For my particular case this means I need to set a macOS minimum target of 10.15 (I can't use boost; and you need filesystem)

In some of our other projects faced with a similar problem outside of PEGTL we have used https://github.com/gulrak/filesystem this implementation of std::filesystem which compiles on macOS back to 10.9 or so.

Before I spend any time on it at all, would you have any interest in using that as a fourth filesystem option? It looks like the cmake and internal filesystem redirects wouldn't be that hard, but before I spend a day or two convincing myself of that want to make sure you don't have some reason not to.

Of course 10.15 is a very old OS so it is usually a reasonable minimum version to support, but I'm making music software and a small number musicians are still on 10.14 since they have 32 bit macOS applications which never got ported. I'd rather use tao json and tell them to upgrade than not; but if we can use ghc::fs instead that would be even better.

Thanks for the great software and for your consideration.

ColinH:
Hi, sorry for the wait, we had to ponder a couple of possibilities, especially since the drive towards a PEGTL 4.0.0 is also driving us towards cleaning out support for older compilers and non-standard libraries.

If you can prepare and test a PR that adds support for this additional filesystem library to the PEGTL 3.x branch we would merge the PR, tag a new PEGTL 3.x version, update the PEGTL in taoJSON to that new version, and tag a new taoJSON beta release.

That still gives us the possibility to go the other way in the PEGTL development branch, e.g. remove support `std::experimental::filesystem` and `boost::filesystem` from the PEGTL 4.0 to end up with only `std::filesystem`.

This will eventually be used in taoJSON, too, but you can use the taoJSON version with the new 3.x PEGTL for as long as you need.

Does this work for you?

baconpaul:
That works for me thank you! Let me get to it next few weeks. 

I’m also happy to maintain a branch and realize audio is probably unique in this osx regard

another option I considered: why does taojson even have a file system dependency? I am using it entirely for in memory communication and not for any file handling. I wonder if a build of the suite with zero file system support as an option may be a more tolerable approach for your forward plans? Happy to consider that also

Thanks for these super libraries also

ColinH:
[...]

PR #350, opened 2023-07-02
Add a CMake option to select no std::filesystem

baconpaul:
Cmake option PEGTL_NO_STD_FILESYSTEM will mean there is *no* implementation of std::filesystem selected at compile time. This will invariably lead to a broken PEGTL if you set it and use file operations, but it is intended to allow a configuration on systems without boost or std::filesystem, which don't need filesystem access but do want memory parsing. This allows taojson memory parsers to compile back to macOS 10.13, for instance.

See discussion in issue #347

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,7 @@ target_include_directories(pegtl INTERFACE
 target_compile_features(pegtl INTERFACE cxx_std_17)
 
 option(PEGTL_USE_BOOST_FILESYSTEM "Override the auto-detection of std::filesystem and use Boost.Filesystem" OFF)
+option(PEGTL_NO_STD_FILESYSTEM "Don't link to any std::filesystem; file operations will not work" OFF)
 
 # Try compiling a test program with std::filesystem or one of its alternatives
 function(check_filesystem_impl FILESYSTEM_HEADER FILESYSTEM_NAMESPACE OPTIONAL_LIBS OUT_RESULT)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -50,6 +50,7 @@ target_include_directories(pegtl INTERFACE
 target_compile_features(pegtl INTERFACE cxx_std_17)
 
 option(PEGTL_USE_BOOST_FILESYSTEM "Override the auto-detection of std::filesystem and use Boost.Filesystem" OFF)
+option(PEGTL_NO_STD_FILESYSTEM "Don't link to any std::filesystem; file operations will not work" OFF)
```
uilianries, line 53:
```suggestion
option(PEGTL_STD_FILESYSTEM "Link to any std::filesystem" ON)
```

I would suggest using positive option name instead of negative to avoid misconfiguration when using negative-negative value. Plus, it would enforce the current state, instead of "adding" a new meaning to the filesystem usage.

baconpaul: Thank you for the feedback. I have updated the pull request accordingly to have a positive option and clearer documentation strings.

ColinH:
@uilianries Any reason from your side to not merge this?

@baconpaul You tested this change locally?

I assume that we can't just set up a CI job with FS disabled because many tests and examples use FS operations and would need to be adapted, which I'd like to avoid for being rather intrusive, like the "no exception" case...

baconpaul:
> @baconpaul You tested this change locally?

Yes with this diff applied to PEGTL under my taoJSON I was able to build my target program for macOS 10.13 (which solves my immediate problem for the subset of my audience still stuck on 10.14).  The diff except for moving the OSX version required from my software (including no longer including json.h and msg pack.h) is already merged [here](https://github.com/surge-synthesizer/shortcircuit-xt/commit/a82aae1a3a2e25c2cab0f9d09534f83ecf5b8cb0).

Thanks so much!

uilianries:
@ColinH I'm fine with this PR, the configuration looks good. I could add a new test to cover this new option on next week (this week I'm totally busy).

uilianries (approved):

baconpaul:
Thank you so much for merging this!

Zero rush, zero pressure, zero expectation, but do you have an idea how long this would take to be released as a dependency of a version of taoJSON?

ColinH:
@uilianries On further thought, this PR, and any work building on it, should be done on the 3.x branch. The main branch for 4.0 will be changed to assume `std::filesystem` and not support any alternatives like `boost::filesystem`.

ColinH:
> Zero rush, zero pressure, zero expectation, but do you have an idea how long this would take to be released as a dependency of a version of taoJSON?

This change needs to be moved to the 3.x branch, and when @uilianries finishes the tests he wants to do we can tag a new version and put that into taoJSON. I don't dare estimate how long that'll take ;-)

baconpaul:
> > Zero rush, zero pressure, zero expectation, but do you have an idea how long this would take to be released as a dependency of a version of taoJSON?
> 
> This change needs to be moved to the 3.x branch, and when @uilianries finishes the tests he wants to do we can tag a new version and put that into taoJSON. I don't dare estimate how long that'll take ;-)

Thanks. My project is t shipping until after summer anyway so I have loads of time and can just fork if needs be too. Super appreciate the help here. 

Thank you

PR #351, opened 2023-07-10
Validate PEGTL tests when no filesystem is configured

uilianries:
I merged the PR #350 to a new branch created from 3.x, and added a new new validation in Actions to check that we don't break tests when running without filesystem.

The filesystem has a specific test `test_filesystem.cpp.in` that should be activated only when `PEGTL_USE_FILESYSTEM` is active. So this new regression test checks in case we break something when changing filesystem.

Related to #350

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -23,7 +23,7 @@ jobs:
           - clang++-12
         build_type: [Debug, Release]
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     env:
       CXX: ${{ matrix.compiler }}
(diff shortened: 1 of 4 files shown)


ColinH:
Great, thanks, you can merge this in to the 3.x branch when you think it's ready, and I'll take care of removing the change and `boost::filesystem` support from the main branch RSN.

uilianries:
@ColinH I see Github Action is failing due missing compilers. I'll take a look on this. It seems like the latest version of Ubuntu distributed by Github does not support older compiler versions.

uilianries:
Clang Tidy failed, but I'll ignore it for now, as it could be customized.

PR #352, opened 2023-07-15
add EOL option for cr_crlf_lf

rowlesmr:
I'm currently writing a parser where the grammar specifies a line termination as `(\r\n?)|\n`. This (I don't think) is currently an option for a valid line ending in PEGTL.

I wrote the following code as an attempt to rectify this. I think I added everything that I needed to, but there is a high chance I didn't.

I added to the tests in [ascii_eol.cpp](https://github.com/taocpp/PEGTL/blob/main/src/test/pegtl/ascii_eol.cpp). I'm assuming that the final argument in `verify_rule` is the number of characters remaining in the input string after a single sucessful match of the rule to the beginning of the string.

Is this something you're interested in?

### DIFF
--- a/include/tao/pegtl/eol.hpp
+++ b/include/tao/pegtl/eol.hpp
@@ -9,6 +9,7 @@
 
 #include "internal/eol.hpp"
 
+#include "internal/cr_crlf_lf_eol.hpp"
 #include "internal/cr_crlf_eol.hpp"
 #include "internal/cr_eol.hpp"
 #include "internal/crlf_eol.hpp"
@@ -28,6 +29,7 @@ namespace TAO_PEGTL_NAMESPACE
          // clang-format off
          struct cr : internal::cr_eol {};
          struct cr_crlf : internal::cr_crlf_eol {};
+         struct cr_crlf_lf : internal::cr_crlf_lf_eol {};
          struct crlf : internal::crlf_eol {};
          struct lf : internal::lf_eol {};
          struct lf_crlf : internal::lf_crlf_eol {};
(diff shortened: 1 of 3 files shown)


ColinH:
Yes, we are interested in this, however there's a reason we haven't added it yet ourselves.

The `eol` classes have a `static constexpr int ch = /* '\r' or '\n' */` data member that is used for the position tracking, i.e. whenever this character is encountered in the input the current line number is incremented by one (and the current column number reset to zero).

At this point you probably understand the problem, within the current framework we can't have an `eol` that can match a single `\r` or a single `\n` because there's no common single character to use for the line counting.

The good news is that my input refactoring experiments change the whole approach to position tracking and end-of-line handling that makes it possible to use an arbitrary PEGTL rule for end-of-lines.

The bad news is that this input refactoring is not at all simple and will take more time to finish; if we're lucky it will be done by the end of next year.

rowlesmr:
> there's a reason we haven't added it yet ourselves. 

I thought it was too easy. 

> The `eol` classes have a `static constexpr int ch = /* '\r' or '\n' */` data member that is used for the position tracking.

Ah. I hadn't figured out what this was for yet. 

> The good news
👍

> The bad news
😪

I think I'll either default to crlf_lf, or make a `line_term` rule. `\r` line endings should be sufficiently rare in my use case that it might not matter, but I would like to comply with the spec. 

At least know there is one project that will use your new work! 


Thanks!

ColinH:
Yes, I'll remember that there is a real-world use case ;-)

Holler if you need any help with your workaround...

PR #361, opened 2024-04-15
Small grammar/wording fix in Actions-and-States.md

gitamohr:


ColinH:
Thank you, that does read better!

PR #368, opened 2024-08-15
add missing include to fix build for some compilers

felixguendling:
On some compilers / standard library combinations (e.g. Mac OS) the build does not work
This adds missing includes and fixes the build.

### DIFF
--- a/include/tao/pegtl/normal.hpp
+++ b/include/tao/pegtl/normal.hpp
@@ -5,6 +5,7 @@
 #ifndef TAO_PEGTL_NORMAL_HPP
 #define TAO_PEGTL_NORMAL_HPP
 
+#include <exception>
 #include <string>
 #include <type_traits>
 #include <utility>
@@ -23,7 +24,6 @@
 #include "demangle.hpp"
 #else
 #include "internal/dependent_false.hpp"
-#include <exception>
 #endif
 
 namespace TAO_PEGTL_NAMESPACE


uilianries:
@felixguendling Hello! Thank you for your contribution. Could you please share the error output that you are trying to fix? When you say MacOS, which compiler and version are you using? Regards!

felixguendling:
`std::throw_with_nested` is part of the `#include <exception>` header which is not included. So that it works with libstdc++ is more a coincidence and might change in the future. With the latest AppleClang version as well as Clang 17 with `-stdlib=libc++` it produces compilation errors that it can't find `std::throw_with_nested´.

felixguendling:
Compiler output on MacOS:
```
/Users/felix/code/oh/deps/PEGTL/include/tao/pegtl/normal.hpp:70:18: error: no member named 'throw_with_nested' in namespace 'std'
            std::throw_with_nested( parse_error( Rule::error_message, am ) );
            ~~~~~^
/Users/felix/code/oh/deps/PEGTL/include/tao/pegtl/normal.hpp:73:18: error: no member named 'throw_with_nested' in namespace 'std'
            std::throw_with_nested( parse_error( "parse error matching " + std::string( demangle< Rule >() ), am ) );
            ~~~~~^
2 errors generated.
```

Version:
```
c++ --version
Apple clang version 15.0.0 (clang-1500.3.9.4)
Target: arm64-apple-darwin23.5.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

PS: I'm using PEGTL to parse OpenStreetMap like opening hours ([example](https://github.com/felixguendling/oh/blob/master/test/test.cc)). It's really simple to use and I didn't expect compilation times to be so fast. Thank you for providing this library!

uilianries:
@felixguendling Thank you for providing such details!

d-frey:
This is IMHO not the correct fix. `<exception>` is included several lines below in the header *iff* `__cpp_exceptions` is set. The PEGTL supports compiling with `-fno-exception` and therefore, you should `#if`-... the parts where `throw_with_nested` is used. CC @ColinH

felixguendling:
Looking at the `#if defined(__cpp_exceptions )` part, it looks like the `#include <exception>` statement has already been there, just in the wrong branch if the preprocessor `#if`?

https://github.com/taocpp/PEGTL/blob/68f446153c5c736035dbee1286e788f904006add/include/tao/pegtl/normal.hpp#L22-L27

d-frey:
Right. I think this is due to the addition of `raise_nested` which was only added after the infrastructure for `-fno-exception` support was already in place. So please update you PR to remove the second guarded inclusion of `<exception>`, as the first global one should now be correct and sufficient.

felixguendling:
I changed it. This looks better now (and still works for me).

d-frey:
@ColinH I assigned it back to you, as the main branch is currently your playground. The PR is OK from my side.

ColinH:
Right, thanks everybody, we can merge this. I'll manually update the `work` branch where the fun is happening.

PR #374, opened 2024-11-27
[infrastructure] Update Codecov upload steps

uilianries:
The current Codecov is not receiving the Code coverage results anymore, the last report is 3 months old: https://app.codecov.io/gh/taocpp/PEGTL/tree/work

The Codecov changed a bit, now they have an integrated Github Action, that can be connected directly to project: https://github.com/marketplace/actions/codecov

The current upload steps are a bit different, not including the new integrated application, but the token used for upload: https://docs.codecov.com/docs/github-2-getting-a-codecov-account-and-uploading-coverage#install-the-github-app-integration

### DIFF
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -28,5 +28,7 @@ jobs:
     - working-directory: build/
       run: ctest --output-on-failure
 
-    - working-directory: build/
-      run: bash <(curl -s https://codecov.io/bash)
+    - name: Upload coverage reports to Codecov with GitHub Action
+      uses: codecov/codecov-action@v4.2.0
+      with:
+        token: ${{ secrets.CODECOV_TOKEN }}


d-frey:
Thanks for working on this! I tried adding a code coverage job to taocpp/taopq, but it failed. Could you move the token from the PEGTL to the organization level (taocpp)? (Given that this would allow all taocpp-projects to use Codecov)

uilianries:
@d-frey That's not possible, not because Github does not support (it does: https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#creating-secrets-for-an-organization), but because Codecov generates different tokens for each project, as fine grain token, so in case one token is compromised, we don't risk all projects. 

I can adjust it right now, but to you understand the steps:

- Enter at https://app.codecov.io/gh/taocpp/taopq/config/general
- Copy the token provided for taopq
- Enter at https://github.com/taocpp/taopq/settings/secrets/actions
- Add a "New Repository Secret"
- Name it as CODECOV_TOKEN and pass the token as its value
- Done!

uilianries:
@d-frey @ColinH The Taopq is working again on the Codecov page: https://app.codecov.io/gh/taocpp/taopq. An incredible coverage of 96.96%!!!

I see other projects are not configured yet, I'll take a look by today during my evening.

PR #377, opened 2025-03-16
Implementing peg2pegtl, based on abnf2pegtl.

redmercury:
This PR implements peg2pegtl based on abnf2pegtl. The input grammar is from the [original PEG paper](https://pdos.csail.mit.edu/~baford/packrat/popl04/peg-popl04.pdf).


redmercury:
I'm a little confused about the clang-tidy errors given that they're in files untouched by this PR.  Any insights into that?

ColinH:
> I'm a little confused about the clang-tidy errors given that they're in files untouched by this PR. Any insights into that?

Sometimes this happens when the jobs switch to a newer version of clang-tidy, I'll look into this, and the rest of the PR as soon as I have some free time.

redmercury:
Thanks.  Ideally I would have added a test that recompiled peg2pegtl using the generated output from generating a grammar from peg.peg, but I'm not quite sure how to achieve that in cmake (it's been a while since I used it).  I did verify manually that replacing the grammar from peg.hpp with generated output generated the same output though.

ColinH:
This PR still appears to contain some Bazel related files?

uilianries:
@redmercury Thank you for your PR. Please,  keep focused only in the proposed feature, do not add a new build system that's not supported in the project already, please, revert the bazel changes and keep only cmake for now. There is a separate discussion about including bazel in project.

redmercury:
Sorry about that, it wasn't my intention, just an oversight when experimenting.  My apologies. should be good now.

redmercury:
@ColinH Are you okay with this commit now I've removed those files?

ColinH:
We're getting there. A couple of things could be improved in the grammar to bring it more in line with best practices. For example the `Comment` rule can be simplified using `tao::pegtl::until<>`, and the final case of `Char` could be simplified with `not_one<>`. Do you want to give it a go?

ColinH:
Please feel free to re-open if you want to continue this.

PR #379, opened 2025-04-12
[infra] Simplify clang-tidy usage

uilianries:
This PR "improves" the current usage of Clang tidy in the CI:

- First, it runs Cmake to obtain exported commands, which can be analyzed by clang-tidy
- Similar to the first, clang-tidy is now running in parallel with xargs -P
- Only errors from project headers are considered critical for failure
- When failing it will mark as failed, but will not break the CI. Which means it is optional, not a blocker

@ColinH @d-frey We could have it as technical debt, what do you think?

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -35,6 +35,7 @@ Checks: >-
   -readability-identifier-length,
   -readability-magic-numbers,
   -readability-static-accessed-through-instance,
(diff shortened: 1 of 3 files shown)


ColinH:
Sounds good, with clang-tidy there's often that one check that you'd like to keep even if it has a false positive...

PR #388, opened 2026-08-12
`linux.yml`: Cover GCC 15 and 16

hartwork:


ColinH:
Very good, last time I checked g++15 wasn't available yet.

PR #393, opened 2026-08-13
[main] Mass-apply clang-format 20.1.8 (fixes #387)

### LINKED ISSUE #387: Need help with fixing the `clang-format` CI?
Hi!

I noticed that the `clang-format` CI [has been red for a long time](https://github.com/taocpp/PEGTL/actions/workflows/clang-format.yml) and continues to make new pull requests red. I was wondering what you know about the current state, if it's just a version mis-match or if mass-reapplication would be needed etc. I would like to understand if anything makes the fix a hard problem and if maybe I can help some way: I consider continuing to enforce auto-formating the right path. What do you think?

Best, Sebastian

hartwork:
Fixes #387

CC @ColinH


src/example/token_input_1.cpp
```diff
@@ -99,7 +99,7 @@ namespace example
 int main()
 {
    const std::vector< example::my_token > v{
-      { example::my_type::beta, "first" },
+      { example::my_type::beta, "first"  },
```
ColinH, line 102: This is strange and not right, same in `token_input_2.cpp`.

include/tao/pegtl/internal/text_input_with_source.hpp
```diff
@@ -131,15 +131,15 @@ namespace TAO_PEGTL_NAMESPACE::internal
...
+         return previous( pos ) - (std::min)( max, pos.column - 1 );
       }
 
       [[nodiscard]] const data_t* end_of_line_or_file( const error_position_t& pos, const std::size_t max = 135 ) const
       {
          static_assert( sizeof( data_t ) == 1 );  // Generalising beyond 1 requires the until to use eol_lazy_peek like in lazy_input.
          using eolof_grammar = until< at< sor< eof, eol_rule > > >;
          const data_t* p = previous( pos );  // TODO: Start earlier?
-         const std::size_t s = ( std::min )( max, std::size_t( this->end() - p ) );
+         const std::size_t s = (std::min)( max, std::size_t( this->end() - p ) );
```
ColinH: This does not fit with the general style.

src/example/json_builder.cpp
```diff
@@ -93,7 +91,8 @@ namespace example
          else {
             return 1;
          }
-      }, v.data );
+      },
+                         v.data );
```
ColinH, line 95: This doesn't make sense.

ColinH (changes requested):
Most changes are fine, but some are not what I want the code to look like and should be removed, of which I flagged a few examples.

include/tao/pegtl/example/open_metrics.hpp
d-frey: The whole grammar in this file should not be formatted, use // clang-format off

src/example/token_input_1.cpp
hartwork: @ColinH that's actually feature "vertical alignment" at work:

screen

It can be disabled via the config file. It it not a bug, as far as I can tell.

include/tao/pegtl/internal/text_input_with_source.hpp
hartwork: @ColinH that's current explicit config `SpacesInCStyleCastParentheses: false` at work. clang-format does as the current config is asking it to:

https://github.com/taocpp/PEGTL/blob/c46edfb554017a95081cbd0a4e711dddfcf8f4e4/.clang-format#L91

src/example/json_builder.cpp
hartwork: @ColinH it's a actually a vertical alignment feature at work:

screen

It's a matter of finding the right knob and seeing if it's other options feel more natural.

include/tao/pegtl/internal/text_input_with_source.hpp
hartwork: @ColinH I have now flipped the option for `true` in a second commit and re-applied mass-formatting as a third. So you can see the effect in isolation.

hartwork:
@ColinH I addressed your findings one by one above. What do you think?

hartwork:
@ColinH I addressed your findings one by one above. What do you think?

include/tao/pegtl/internal/text_input_with_source.hpp
ColinH: So clang-format is mis-interpreting this as a cast, got it.

src/example/json_builder.cpp
ColinH: To look nice lambda expressions as function arguments probably need separate formatting. Swift does this better.

include/tao/pegtl/internal/peek_member.hpp
```diff
@@ -35,8 +35,8 @@ namespace TAO_PEGTL_NAMESPACE::internal
 
    // For data members of type 'T*' or 'const T*' -- will this case ever be used?
 
-   template< typename C, typename T, T* C::*P >
-   struct peek_member_impl< T* C::*, P, std::enable_if_t< std::is_member_object_pointer_v< T* C::* > > >
+   template< typename C, typename T, T* C::* P >
+   struct peek_member_impl< T * C::*, P, std::enable_if_t< std::is_member_object_pointer_v< T * C::* > > >
```
ColinH, line 39: This is where clang-format seems to think the `T *` is "T times" instead of "T is a pointer".

include/tao/pegtl/example/open_metrics.hpp
```diff
@@ -20,82 +20,139 @@ namespace TAO_PEGTL_NAMESPACE::open_metrics
...
-
-   struct exemplar_labels : seq< OPEN, list_opt< exemplar_label, COMMA >, CLOSE > {};
-   struct sample_labels : seq< OPEN, list_opt< sample_label, COMMA >, CLOSE > {};
-
-   struct exemplar : seq< HASH, SP, exemplar_labels, SP, exemplar_number, opt< SP, exemplar_timestamp > > {};
-   struct sample : seq< sample_metric_name, opt< sample_labels >, SP, sample_number, opt< SP, sample_timestamp >, opt< SP, exemplar >, eol > {};
-
-   struct content : sor< metric_descriptor, sample > {};
-   struct exposition : until< trailer, content > {};
+   struct BS : one< '\\' >
```
ColinH, line 23: This reformatting of grammars needs to be suppressed with `// clang-format off`

ColinH:
Most of the changes that I looked at are good, for the rest see other comments.

include/tao/pegtl/internal/text_input_with_source.hpp
hartwork: Resolving, please unresolve as needed…

include/tao/pegtl/internal/peek_member.hpp
hartwork: @ColinH potentially, yes. I guess your best bet weould be to check for the behavior with clang-format 23.x.x and report a bug with the LLVM team if you consider it broken with 23.x.x as well. I'm reading your reply as "non-critical, can remain as-is. Resolving, please unresolve as needed…

src/example/json_builder.cpp
hartwork: I am reading that as "non-critical, okay as-is", resolving. Please unresolved as needed…

src/example/token_input_1.cpp
hartwork: Resolving, please unresolve as needed…

include/tao/pegtl/example/open_metrics.hpp
hartwork: @ColinH I can wrap things with `clang-format off` but you using plural fro grammars makes me think that I'll need a definition of which blocks in code you'd want me to exclude. Is file `open_metrics.hpp` line 23 to 98 the only grammar to be exclude? If not, what would be the complete list?

hartwork:
@ColinH I'll be honest with you, my patience with and interest in this pull request is exhausted and I'll close it. Fixing clang-format CI is trivial, and in my view first getting CI green with any formatting and then tuning formatter config further in a second pull request is the way I would go, and treat it as two orthogonal problems. I am not needed for that tuning. Thanks for your understanding :pray:

